PR #1, opened 2015-10-25
Added GPLv3 License

enochtsang:
I added a copy of the AGPLv3 license and put a permissions statement at the top of every source file.

### DIFF
--- a/src/BatteryData.cpp
+++ b/src/BatteryData.cpp
@@ -1,3 +1,30 @@
+/** 
+ *  University of Calgary Solar Car Team New Recruit Exercise
+ *  Copyright (C) 2015 University of Calgary Solar Car Team
+ *
+ *  This file is part of the 
+ *  University of Calgary Solar Car Team New Recruit Exercise
+ *
+ *  The University of Calgary Solar Car Team New Recruit Exercise 
+ *  is free software: you can redistribute it and/or modify it 
+ *  under the terms of the GNU General Public License as 
+ *  published by the Free Software Foundation, either version 3 
+ *  of the License, or (at your option) any later version.
+ *
+ *  The University of Calgary Solar Car Team New Recruit Exercise 
+ *  is distributed in the hope that it will be useful, but WITHOUT 
+ *  ANY WARRANTY; without even the implied warranty of MERCHANTABILITY 
+ *  or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General 
+ *  Public License for more details.
+ *
+ *  You should have received a copy of the GNU General 
+ *  Public License along with the University of Calgary Solar Car 
+ *  Team New Recruit Exercise. If not, see 
+ *  <http://www.gnu.org/licenses/>.
+ *
+ *  For further contact, email <software@calgarysolarcar.ca>
+ */
+ 
 #include "BatteryData.h"
 
 BatteryData::BatteryData()
--- a/src/BatteryData.h
+++ b/src/BatteryData.h
@@ -1,3 +1,30 @@
+/** 
+ *  University of Calgary Solar Car Team New Recruit Exercise
+ *  Copyright (C) 2015 University of Calgary Solar Car Team
+ *
+ *  This file is part of the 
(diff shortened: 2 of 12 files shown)


JordanHeinrichs:
To be honest I am not a fan of including the preamble in all the source files. I think it is needless clutter and just added work whenever we create new files. 

Looking at http://programmers.stackexchange.com/questions/125836/do-you-have-to-include-a-license-notice-with-every-source-file they recommended it but it is not required. 

We don't have any competitive or financial reason to make the license any more explicit then it actually is. Having the license file within the repo main fulfills our obligation that we have when using the open source libraries that we do and anything more is unneeded.

I would be for closing and not merging this but I want to here your opinion.

enochtsang:
I would still like to include at the top of every source file. If it is an extra bit of work to add that at the top of every new source file, then we can leave that to me in every pull request. This licensing stuff is boring of course, so we will keep people developing and I will handle the boring stuff with licenses. 

You're right that we don't have much to worry due to lack of financial reason or competitiveness, but I think it's a small thing to do to keep us safe.

JordanHeinrichs:
What do we have to be safe from? What could happen if we don't?

enochtsang:
Ok safe was the wrong word. To be honest, I just thought having an explicit license was a good idea but I wasn't sure why. But I did some reading and people make a good point that developers will be hesitant to use our software if we don't properly include a license, this defeats the whole purpose of us wanting to open-source everything on the team anyways.

JordanHeinrichs:
I thought the whole purpose of open sourcing it was just so that potential employers can see the project and not other developers using the software. 

But I also did some reading and they do make a good point including it although I really don't think there is any risk of individual source files getting separated from the project and used maliciously. We could shorten it to exactly what is in http://www.gnu.org/copyleft/gpl.html as there really doesn't need to be the team's email in every single file. If you are opposed to shorting it then I could merge it in.

I suppose I am just more of a BSD or MIT fan, or even WTFPL (although I would never actually use this license...).

enochtsang:
I think if we shorten it, it defeats the whole purpose of wanting to not include it in every file, since we have to add something to every file anyways. I was looking at MIT and I think that would also be acceptable, I just liked GPL a bit better. So even though GPL is a bit more work, I wanted to use that instead since it was a bit better. I've also read BSD requires editing from project to project, and I don't really wanna do that. 

And lol, the WTFPL license, let's use that haha.

Also you make a good point about how originally it was supposed to be open source so we can show employers, but I think with some of the good things we're doing I also want to make it easy for other teams to use our code.

mpminardi:
I think the decision ultimately depends on how we want / see people using the code. I can see certain projects (i.e, OnBoard-Media-Control) having individual files used if the code is used at all. If we do not include something in each file, there may be a chance of some grey area for licensing (although having a LICENSE / COPYRIGHT file in the top level should be fairly explicit). We could have a top level license file, and then something to the effect of:

Copyright 2015 University of Calgary Solar Car Team. Licensed under GNU Affero General Public License (AGPL). See the LICENSE file at the top-level directory or <website with version of the license> for details.

in each file, which explicitly lays out the licensing intent / directs to the full license for details.

bseto:
I have a question: What if some code in our repository is from a stackoverflow answer?
https://github.com/UCSolarCarTeam/Schulich-Delta-OnBoard-Media-Control/blob/master/include/threadClass.hpp

In the above class, the class is pretty much a straight up copy of the answer in a stackoverflow reply. Do we have to be more formal regarding where the code came from?

enochtsang:
I was actually wondering about that too, when I went over it I just thought we could use it. But since you asked I actually looked it up, sounds complicated. I might ask you to rewrite. I will also ask Jonah to take a closer look for us.

JordanHeinrichs:
If you decided to go with AGPL make sure to update the license file since currently the GPLv3 resides within the main directory of the repo. I would recommend just updating the headers to the GPL one but that is up to you.

After that I will merge it.

enochtsang:
The headers are GPL, can you merge? :3

JordanHeinrichs:
Resolved pull request. GPLv3 headers are now added.

PR #3, opened 2015-10-31
Ben

benji373:
Completed the assignment.  Program outputs the time until the battery is depleted or charged based on whether or not the battery is charging or depleting.

PR #5, opened 2015-11-07
second pull request for recruit training

benji373:
Commits were reviewed and code was fixed.  variable names were changed to be more logical.  Variables that were only used in one function were moved out of the header.  The calculations to find the time remaining on the charge were moved into the calculate time remaining function.


enochtsang:
Refer to pull request #3

PR #7, opened 2015-11-07
Davis Roman

DavisRomn:
Made changes to code and more or less finished up.
-Edited State of Charge Service
-Edited Headers
-Edited Display


enochtsang:
This pull request is the wrong way around! At the top you can see, 
"wants to merge 2 commits into Davis_Roman from master"
it should be 
"wants to merge n commits into master from Davis_Roman"

PR #8, opened 2015-11-07
Bill luu

bill-luu:


enochtsang:
Stage 1: Is the purpose of the request sound?

Yes, stage 1 finished.
Starting stage 2: Code structure.

src/LogFileReader.cpp
```diff
@@ -75,11 +77,23 @@ bool LogFileReader::readAll(const QString& fileName)
...
+}
 
-    // This is here to the compiler happy. Otherwise the compile
-    // will have an error warning about an unused variable. Remove this
-    // when you use it.
-    Q_UNUSED(line);
-    Q_UNUSED(batteryData);
+QString timeString= sections.at(0);
+//Converts timeString to time for battery data
+batteryData.time= QTime::fromString(timeString,"hh:mm:ss.zzz");
```
enochtsang: This file is very clean, well done. I would note that in general writing strings in non print functions, as well as straight numbers in functions calls is poor structure. Normally, you would either derive such a value from somewhere else, or you would give it a descriptive name as a constant variable somewhere.

In this case, the "hh:mm:ss.zzz" constant, since it can't be derived elsewhere, it should made as a constant. Normally how we do constants in C++ on the solar car team is adding a namespace at the top. So the code would be something like:

namespace 
{
    const QString SOME_DESCRIPTIVE_NAME = "hh:mm:ss.zzz";
}
I would suggest something similar for the line QStringList splitLine = line.split(", "); as well as the 3.

src/LogFileReader.cpp
```diff
@@ -29,9 +29,10 @@
 #include <QFile>
 #include <QString>
 #include <QTextStream>
-
 #include "LogFileReader.h"
-
+#include <iostream>
```
enochtsang: This header is not being used and should be removed.

src/BatteryStateOfChargeService.h
```diff
@@ -45,5 +45,12 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
 
+    double AmpHoursUsed_;
+    double inCurrent_;
+    double newCurrent_=0;
+    double sumCurrent_=0;
+    bool firstRun_=true;
```
enochtsang: Also, all initializations should be done in the `.cpp` file. It gives a central point for readers to know how variables are initialized.This issue concerns lines 50-52, these were listed as warnings when you compiled, for some newer versions it comes up as a compiler error instead of a warning. To fix this:

```c++
 , newCurrent (0)
{
...
}
```
is the syntax to use. 
```c++
 {
   newCurrent = 0;
   ...
  }
```
is also valid but the other way is superior because of performance reasons (if you are curious https://isocpp.org/wiki/faq/ctors#init-lists)

src/BatteryStateOfChargeService.cpp
```diff
@@ -35,29 +39,58 @@ namespace
...
+
+    int mschangeTime=abs(initialTime_.msecsTo(currentTime_));
+    double changeTime=mschangeTime*2.77778e-7;
+
+    double avgCurrent=(newCurrent_+inCurrent_)/2;
+    double pointAverageAH= avgCurrent*changeTime;
+
+    AmpHoursUsed_=AmpHoursUsed_-pointAverageAH;
+    }
+    firstRun_=false;
```
enochtsang: Since this line of code `firstRun_ = false;` only needs to be run the first time through, it would be better to include it in the `if` statement. Having it outside implies that some other parts of the code could edit this variable to `true`, which is not true. It also will increase the speed of the program a teeny tiny bit (this last point is not really important since speed is not really an issue with this program).

src/BatteryStateOfChargeService.cpp
```diff
@@ -35,29 +39,58 @@ namespace
...
+    inCurrent_=newCurrent_;
+    initialTime_=currentTime_;
+
+    newCurrent_=batteryData.current;
+    currentTime_=batteryData.time;
+
+    if(firstRun_!=true){
+
+    int mschangeTime=abs(initialTime_.msecsTo(currentTime_));
+    double changeTime=mschangeTime*2.77778e-7;
```
enochtsang: This value `2.77778e-7` is something that we usually call a "magic number", it's a number that does something but has no context as to what it actually does. This number should be abstracted into a constant like how I described in `LogFileReader.cpp`.

src/BatteryStateOfChargeService.cpp
```diff
@@ -35,29 +39,58 @@ namespace
...
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+  double timeLeft;
+  timeLeft=(AmpHoursUsed_/newCurrent_);
+  timeLeft=qAbs(timeLeft);
+  timeLeft=timeLeft/2.77778e-7;
+
+  QTime base(0,0,0);
```
enochtsang: In the documentation of `QTime`, the constructor only requires two values, so `Base(0, 0)` would suffice.

src/BatteryStateOfChargeService.cpp
```diff
@@ -35,29 +39,58 @@ namespace
...
+  QTime timeOutput= base.addMSecs(timeLeft);
```
enochtsang: This new variable `timeOutput` is not really necessary, since `addMSecs` is a member function, using `base.addMSecs` will change base itself. I would suggest replacing this line with:
```c++
return base.addMSecs(timeLeft);
```

src/BatteryStateOfChargeService.h
```diff
@@ -45,5 +45,12 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
 
+    double AmpHoursUsed_;
+    double inCurrent_;
```
enochtsang: `inCurrent_` is only used in one function and doesn't really need to stored between function calls, it would be better suited as a variable in `addData` rather than a private member.

src/BatteryStateOfChargeService.h
```diff
@@ -45,5 +45,12 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
+    double newCurrent_=0;
+    double sumCurrent_=0;
+    bool firstRun_=true;
+    QTime initialTime_;
```
enochtsang: Same comments as with `inCurrent_`. In particular though, `initial` is not very accurate since you are changing it every time you `addData` so it is only the "initial time" the first time `addData` is called.

src/BatteryStateOfChargeService.cpp
```diff
@@ -26,6 +26,10 @@
  */
 
 #include "BatteryStateOfChargeService.h"
+#include "BatteryData.h"
+#include <QTextStream>
+#include <QDebug>
+#include <QFile>
```
enochtsang: Lines 30-32 are not being used in this file, so they should be removed to not confuse a reader as to their purpose.

src/LogFileReader.cpp
```diff
@@ -29,9 +29,10 @@
...
+#include <QTime>
+using namespace std;
```
enochtsang: This line is also not necessary, but I'd also like to note that the directive `using namespace ...;` in general is frowned upon because it creates ambiguity for where some words come from.

src/BatteryStateDisplayService.cpp
```diff
@@ -51,10 +51,16 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
...
+    if(batteryStateOfChargeService_.isCharging())
+        s=" Time Until Charge: ";
+
 
     QTextStream(stdout) << "Voltage: " << batteryData.voltage
-        << " Current: " << batteryData.current
-        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed() << endl;
+       << " Current: " << batteryData.current
+        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed()
+        <<s<<batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString("hh:mm:ss.zzz")
```
enochtsang: Same thing as in `LogFileReader.cpp` for the `"hh:mm:ss.zzz"` constant!

src/main.cpp
```diff
@@ -28,7 +28,7 @@
 #include <QCommandLineParser>
 #include <QCoreApplication>
 #include <QTimer>
-
+#include <iostream>
```
enochtsang: This isn't really necessary, includes should be per file. The main doesn't require `<iostream>`.

src/BatteryStateOfChargeService.cpp
```diff
@@ -26,38 +26,70 @@
...
+  return base.addMSecs(timeLeft);
 }
 
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+ //Setting Values
+  double inCurrent_=newCurrent_;
+   QTime previous_Time=currentTime_;
```
enochtsang: Since both `inCurrent` and `previous_Time` are both "previous" values, it'd be more clear and make more sense to have the same prefix for both these variables.

src/BatteryStateOfChargeService.cpp
```diff
@@ -26,38 +26,70 @@
...
+
+    newCurrent_=batteryData.current;
+    currentTime_=batteryData.time;
```
enochtsang: Same thing as before, it'd be better if `newCurrent_` and `currentTime_` had the same prefix.

src/BatteryStateOfChargeService.cpp
```diff
@@ -26,38 +26,70 @@
...
+
+    if(firstRun_!=true){
+
+    int mschangeTime=abs(previous_Time.msecsTo(currentTime_));
```
enochtsang: `mschangeTime` could use a better name, `msSinceLastTime` I think would be pretty suitable.

src/BatteryStateOfChargeService.cpp
```diff
@@ -26,38 +26,70 @@
...
+    double changeTime=mschangeTime*MSECS_TO_HOURS;
```
enochtsang: Same thing with this variable :)

src/BatteryStateOfChargeService.cpp
```diff
@@ -26,38 +26,70 @@
...
+
+    double avgCurrent=(newCurrent_+inCurrent_)/2;
+    double pointAverageAH= avgCurrent*changeTime;
```
enochtsang: The `AH` should be expanded and not left as an abbreviation.

src/BatteryStateOfChargeService.cpp
```diff
@@ -26,38 +26,70 @@
...
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+ //Setting Values
+  double inCurrent_=newCurrent_;
+   QTime previous_Time=currentTime_;
+
+    newCurrent_=batteryData.current;
+    currentTime_=batteryData.time;
+
+    if(firstRun_!=true){
```
enochtsang: So it's generally better to write an `if` statement for special cases, instead of writing it for the default. So you should move all the logic inside this `if` statement outside, and move the special case logic into the `if` statement.

src/BatteryStateOfChargeService.h
```diff
@@ -45,5 +45,10 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
 
+    double AmpHoursUsed_;
+    double newCurrent_;
+    double sumCurrent_;
```
enochtsang: This variable is not really being used for anything productive, it's just being updated, should probably delete it since it's not being used.

src/BatteryStateOfChargeService.h
```diff
@@ -45,5 +45,10 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
+    bool firstRun_=true;
```
enochtsang: Initialize this value in the constructor, might not compile in some compilers.

src/BatteryStateOfChargeService.cpp
```diff
@@ -26,38 +26,70 @@
...
+    return true;
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+  double timeLeft;
+  timeLeft=(AmpHoursUsed_/newCurrent_);
+  timeLeft=qAbs(timeLeft);
+  timeLeft=timeLeft/MSECS_TO_HOURS;
```
enochtsang: I'd recommend reducing these 4 lines into one line.

src/BatteryStateDisplayService.cpp
```diff
@@ -51,10 +53,16 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
 void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& batteryData)
 {
     batteryStateOfChargeService_.addData(batteryData);
+    QString s= " Time Until Depletion";
```
enochtsang: `s` could use a better name.

src/BatteryStateDisplayService.cpp
```diff
@@ -31,7 +31,9 @@
 #include "BatteryStateDisplayService.h"
 #include "I_BatteryDataSource.h"
 #include "I_BatteryStateOfChargeService.h"
-
+namespace{
+const QString STRING_TIME_CONVERSION= "hh:mm:ss.zzz";
```
enochtsang: I would suggest the name `STRING_TIME_FORMAT`

src/BatteryStateOfChargeService.cpp
```diff
@@ -26,38 +26,67 @@
...
+    if(firstRun!=false)
+        firstRun=false;
+    else
+    {
+
+            int msSinceLastTime=abs(previous_Time.msecsTo(presentTime_));
+            double HoursSinceLastTime=msSinceLastTime*MSECS_TO_HOURS;
+
+            double avgCurrent=(present_Current_+previousCurrent_)/2;
+            double pointAverageAmpHours= avgCurrent*HoursSinceLastTime;
```
enochtsang: Change the `pointAverage` to something more layman terms.

src/LogFileReader.cpp
```diff
@@ -29,9 +29,14 @@
...
 #include <QString>
 #include <QTextStream>
-
 #include "LogFileReader.h"
-
+#include <QTime>
+namespace
+{
+const QString STRING_TIME_CONVERSION= "hh:mm:ss.zzz";
+const QString SPLIT= ", ";
```
enochtsang: I would suggest using the word `DELIMITER` instead of `SPLIT`

src/LogFileReader.cpp
```diff
@@ -75,11 +81,23 @@ bool LogFileReader::readAll(const QString& fileName)
...
+QString timeString= sections.at(0);
+//Converts timeString to time for battery data
+batteryData.time= QTime::fromString(timeString,STRING_TIME_CONVERSION);
+//Takes voltage
+bool voltageOK;
+batteryData.voltage=sections.at(1).toDouble(&voltageOK);
+//Takes curent
+bool currentOK;
+batteryData.current=sections.at(2).toDouble(&currentOK);
+if(!voltageOK || !currentOK ||!batteryData.time.isValid())
```
enochtsang: I would rewrite this section to be 
`if(!(voltageOK && currentOK && batteryData.time.isValid()))`

enochtsang:
Stage 2 of Pull Request Finished
Stage 3: Bugs and Glitches

enochtsang:
When you print 
`Voltage: 2.95314 Current: 4.0337 Total Ah used: 53.0833 Time Until Depletion13:09:35.768`
should match the format of 
`Voltage: 2.92759 Current: -81.7903 Total Ah used: 55.149 Time Until Charge: 00:40:27.384`

enochtsang:
Stage 3 of Pull Request Finished
Stage 4: Coding Style

enochtsang:
@sam-ao

enochtsang:
In general, you should look for these things to fix

All indentation should be 4 spaces
Opening brace should be on the next line
Variable names other than constants should be camelCase

src/LogFileReader.cpp
```diff
@@ -29,9 +29,14 @@
 #include <QFile>
 #include <QString>
 #include <QTextStream>
-
 #include "LogFileReader.h"
-
+#include <QTime>
+namespace
```
enochtsang: Includes should have one line separating them from the rest of the code.

src/LogFileReader.cpp
```diff
@@ -29,9 +29,14 @@
...
+{
+    const QString STRING_TIME_FORMAT= "hh:mm:ss.zzz";
+    const QString BatData_Delimiter= ", ";
+    const int COLUMNS = 3;
+}
 LogFileReader::LogFileReader()
```
enochtsang: A newline could be used here as well.

enochtsang:
Remove the `TODO` comments
Operators like `=` and `+` should have a space surrounding them.

src/LogFileReader.cpp
```diff
@@ -75,11 +79,23 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    {
+        return false;
+    }
+    QString timeString= sections.at(0);
+    batteryData.time= QTime::fromString(timeString,STRING_TIME_FORMAT);
+    bool voltageOK;
+    batteryData.voltage=sections.at(1).toDouble(&voltageOK);
+    bool currentOK;
+    batteryData.current=sections.at(2).toDouble(&currentOK);
+    if(voltageOK && currentOK && batteryData.time.isValid())
```
enochtsang: This section here could use some line breaks to separate ideas.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +45,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
 
 double BatteryStateOfChargeService::totalAmpHoursUsed() const
 {
-    return 0.0;
+    return (amphoursUsed_);
```
enochtsang: The brackets are not necessary and create unnecessary noise.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +45,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    {
+        return true;
+    }
 }
-
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    double timeLeft;
+    timeLeft=(qAbs(amphoursUsed_/present_Current_))/MSECS_TO_HOURS;
```
enochtsang: The brackets around `qAbs(amphoursUsed_/present_Current_)` aren't necessary

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +45,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+        firstRun_=false;
+    }
+    else
+    {
+        int msSinceLastTime=abs(previous_Time.msecsTo(presentTime_));
+        double hoursSinceLastTime=msSinceLastTime*MSECS_TO_HOURS;
+        double avgCurrent=(present_Current_+previousCurrent_)/2;
+        double ampHoursBetweenPresent_PreviousTime= avgCurrent*hoursSinceLastTime;
+        amphoursUsed_=amphoursUsed_-ampHoursBetweenPresent_PreviousTime;
+    }
```
enochtsang: This section could use some line breaks separating ideas.

src/BatteryStateOfChargeService.h
```diff
@@ -45,5 +45,9 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
 
+    double amphoursUsed_;
```
enochtsang: Proper camelCase would dictate that `amphoursUsed_` should be `ampHoursUsed_`

src/BatteryStateOfChargeService.h
```diff
@@ -45,5 +45,9 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
+    double present_Current_;
```
enochtsang: The underscore after `present` is not proper camelCase.

src/BatteryStateDisplayService.cpp
```diff
@@ -51,10 +55,16 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
...
+        Charge_Depletion_Display=" Time Until Charge: ";
+    }
 
-    QTextStream(stdout) << "Voltage: " << batteryData.voltage
-        << " Current: " << batteryData.current
-        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed() << endl;
 
-    // TODO Print out time till it is depleted or charged.
+    QTextStream(stdout) << "Voltage: " << batteryData.voltage
+       << " Current: " << batteryData.current
```
enochtsang: Match the indentation for this line.

src/BatteryStateDisplayService.cpp
```diff
@@ -51,10 +55,16 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
 void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& batteryData)
 {
     batteryStateOfChargeService_.addData(batteryData);
+    QString Charge_Depletion_Display=" Time Until Depletion: ";
```
enochtsang: This variable would be better written as `chargeDepletionDisplay`

src/BatteryStateDisplayService.cpp
```diff
@@ -51,10 +55,16 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
...
 
-    QTextStream(stdout) << "Voltage: " << batteryData.voltage
-        << " Current: " << batteryData.current
-        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed() << endl;
 
-    // TODO Print out time till it is depleted or charged.
+    QTextStream(stdout) << "Voltage: " << batteryData.voltage
+       << " Current: " << batteryData.current
+        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed()
+        <<Charge_Depletion_Display<<batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString(STRING_TIME_FORMAT)
```
enochtsang: This line is quite long, consider breaking it up into a couple lines?

enochtsang:
Stage 4 complete, good to go!

PR #9, opened 2015-11-07
Aidan

revaniite:


src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +44,65 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+    if (prTime_.isNull())
+    {
+        totalAH_=BATTERY_AMP_HOUR_CAPACITY*initialStateOfChargePercent_/100;
+        prTime_=batteryData.time;
+    }
+    deltaT_=prTime_.msecsTo(batteryData.time);
+    deltaC_=(batteryData.current+preC_)/2;
+    amph_=deltaT_*deltaC_/3600000;
```
mpminardi: Make sure you define any constants that you use for calculations (like the 3600000 here). Using a constant with a descriptive name instead of a "magic number" will help make the code easier to understand and more maintainable.

src/LogFileReader.cpp
```diff
@@ -74,12 +74,19 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    double bdc = line.section(", ",2,2, flag).toDouble();
+    if (bdt.isNull() || !bdv || !bdc){
+        qDebug() << "\nError reading from data"<<endl;
+        return false;
+    }
+    else {
+        batteryData.time=bdt;
+        batteryData.current=bdc;
+        batteryData.voltage=bdv;
+    }
```
enochtsang: Alot of your variables here are heavily abbreviated and makes it hard to understand what the purpose of the variables are. In general (with some exceptions) you should avoid abbreviations when naming variables. Some notable exceptions are `num` for number, and `i` and `j` in `for` loops.

src/LogFileReader.cpp
```diff
@@ -74,12 +74,19 @@ bool LogFileReader::readAll(const QString& fileName)
...
-    // TODO implement this first
-
-    // This is here to the compiler happy. Otherwise the compile
-    // will have an error warning about an unused variable. Remove this
-    // when you use it.
-    Q_UNUSED(line);
-    Q_UNUSED(batteryData);
+    QString::SectionFlag flag = QString::SectionSkipEmpty;
+    QString _bdt = line.section(",",0,0, flag);
+    QTime bdt = QTime::fromString(_bdt, "hh:mm:ss.zzz");
```
enochtsang: In general writing strings in non print functions, as well as straight numbers in functions calls is poor structure. Normally, you would either derive such a value from somewhere else, or you would give it a descriptive name as a constant variable somewhere. 

In this case, the `"hh:mm:ss.zzz"` constant, since it can't be derived elsewhere, it should made as a constant. Normally how we do constants in C++ on the solar car team is adding a namespace at the top. So the code would be something like:
```c++
namespace 
{
    const QString SOME_DESCRIPTIVE_NAME = "hh:mm:ss.zzz";
}
```
I would suggest something similar for the line `QString _bdt = line.section(",",0,0, flag);`

src/LogFileReader.cpp
```diff
@@ -74,12 +74,19 @@ bool LogFileReader::readAll(const QString& fileName)
...
 bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) const
 {
-    // TODO implement this first
-
-    // This is here to the compiler happy. Otherwise the compile
-    // will have an error warning about an unused variable. Remove this
-    // when you use it.
-    Q_UNUSED(line);
-    Q_UNUSED(batteryData);
+    QString::SectionFlag flag = QString::SectionSkipEmpty;
```
enochtsang: I would be wary of this line, since skipping sections that are empty could be dangerous since you know exactly what the first second and third sections should be. If there is an empty section, there is a problem and you should know and indicate that there is an issue.

src/LogFileReader.cpp
```diff
@@ -74,12 +74,19 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    if (bdt.isNull() || !bdv || !bdc){
+        qDebug() << "\nError reading from data"<<endl;
+        return false;
+    }
+    else {
+        batteryData.time=bdt;
+        batteryData.current=bdc;
+        batteryData.voltage=bdv;
+    }
     return true;
```
enochtsang: this `return true` would fit better in the `else` statement since that is what happens when the `if` statement is false.

src/BatteryStateOfChargeService.h
```diff
@@ -38,12 +38,21 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
+    QTime prTime_;
+    double deltaC_;
+    double preC_;
+    double deltaT_;
+    double amph_;
+    double totalAH_;
+    double isCharging_;
+    double deltaAH_;
+    double chargeTime_;
+    int i;
```
enochtsang: Same thing as before about avoiding abbreviations :)

src/BatteryStateOfChargeService.h
```diff
@@ -38,12 +38,21 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     double totalAmpHoursUsed() const;
     bool isCharging() const;
     QTime timeWhenChargedOrDepleted() const;
+    QTime time;
```
enochtsang: There are very few cases when a member is going to be public, this QTime time should be put as a private member, and if it is really necessary that it is accessed, a getter function should be used instead.

src/BatteryStateDisplayService.cpp
```diff
@@ -51,10 +51,19 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
 void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& batteryData)
 {
     batteryStateOfChargeService_.addData(batteryData);
+    bool charging = batteryStateOfChargeService_.isCharging();
+    QString time=batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString("hh:mm:ss.zzz");
```
enochtsang: Same thing here with having constants be defined in a namespace.

src/BatteryStateDisplayService.cpp
```diff
@@ -51,10 +51,19 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
...
+    QString ans;
```
enochtsang: This `ans` variable could use a better name, "answer to what?"

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +44,65 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 {
-    return false;
+    return isCharging_;
 }
 
+
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    QTime ntime(0,0,0,0);
```
enochtsang: This `ntime` variable could use a better name, the `n` is not very descriptive.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +44,65 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+        isCharging_=false;
+        deltaAH_=totalAH_;
+    }
+    deltaAH_ /= deltaC_;
+
+    int hours = deltaAH_;
+    int min = (deltaAH_-hours)*60;
+    int sec =(deltaAH_-hours-(min/60))*60;
+    int ms = (deltaAH_-hours-min/60-sec/3600)*1000;
+    time.setHMS(hours, min, sec, ms);
```
enochtsang: This section could be done cleaner by using the `addMSecsTo` function, and adding milliseconds to a `QTime(0, 0)` object.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +44,65 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 
+
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    QTime ntime(0,0,0,0);
+    if (time.isNull())
+        return ntime;
+    else
+        return time;
```
enochtsang: You should consider moving some of the logic in `addData` into this function, `addData` is quite full right now and some of the logic done for this function is done in `addData`.

src/LogFileReader.cpp
```diff
@@ -32,6 +32,13 @@
 
 #include "LogFileReader.h"
 
+namespace
+{
+    const QString TIME_FORMAT = "hh:mm:ss.zzz";
+    QString TEMP_TIME;
```
enochtsang: The namespace should be reserved for constants, so this variable would better off being initialized in the function it's being used in.

src/LogFileReader.cpp
```diff
@@ -74,12 +81,18 @@ bool LogFileReader::readAll(const QString& fileName)
...
-    // This is here to the compiler happy. Otherwise the compile
-    // will have an error warning about an unused variable. Remove this
-    // when you use it.
-    Q_UNUSED(line);
-    Q_UNUSED(batteryData);
-    return true;
+    TEMP_TIME = line.section(",",0,0);
+    QTime bdt = QTime::fromString(TEMP_TIME, TIME_FORMAT);
+    double bdv= line.section(", ",1,1).toDouble();
+    double bdc = line.section(", ",2,2).toDouble();
```
enochtsang: Generally it's poor practice to abbreviate names, have a longer descriptive name is much better than having a short one for readability reasons. Just expanding them would be pretty good, so `bdt --> batteryDataTime` and so forth.

src/LogFileReader.cpp
```diff
@@ -74,12 +81,18 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    if (bdt.isNull() || !bdv || !bdc){
+        qDebug() << "\nError reading from data"<<endl;
```
enochtsang: This debug while a good idea, is not really necessary since the this function is not in charge of telling the user whether there was a problem or not, it just has to tell the calling function whether there was a problem or not. So just return false is enough :)

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,55 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
 
 double BatteryStateOfChargeService::totalAmpHoursUsed() const
 {
-    return 0.0;
+    if (totalAH_>BATTERY_AMP_HOUR_CAPACITY)
+        return 0;
```
enochtsang: While not necessary, this is good to consider this case!

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,55 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+        totalAH_=BATTERY_AMP_HOUR_CAPACITY*initialStateOfChargePercent_/100;
+        previousTime_=batteryData.time;
+    }
+    zero_time.setHMS(0,0,0,0);
+    deltaTime_=previousTime_.msecsTo(batteryData.time)/HOUR_TO_SEC;
+    deltaCurrent_=(batteryData.current+previousCurrent_)/2;
+    amph_=deltaTime_*deltaCurrent_;
+    totalAH_+=amph_;
+
+    if (batteryData.current < 0){
```
enochtsang: In general, `null` has a different meaning than `0` for objects, so a better name for this variable would be `zeroTime`.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,55 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 {
-    return false;
+    return isCharging_;
 }
 
+
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    QTime null_time(0,0,0,0);
```
enochtsang: In general, `null` has a different meaning than `0` for objects, so a better name for this variable would be `zeroTime`.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,55 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+        return null_time;
+    else
+        return return_time;
 }
 
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+    if (previousTime_.isNull())
```
enochtsang: Really good! Out of all the times I've seen how people checked for the first time this function has been ran, this is probably the best solution (better than my solution! I didn't do this).

src/BatteryStateOfChargeService.h
```diff
@@ -42,8 +42,17 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     void addData(const BatteryData& batteryData);
 
 private:
-    // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
 
-    // Recommend adding necessary variables here
+    QTime previousTime_;
+    QTime return_time;
+    QTime zero_time;
```
enochtsang: I would not recommend using a private member for this. I'd actually recommend using just a zero constructor `QTime(0, 0)` whenever you need a `zero_time`. This is because `QTime(0, 0)` is just as understandable as `zero_time` and it'll make for less code that needs to be read since you don't need a private member for it.

src/BatteryStateOfChargeService.h
```diff
@@ -42,8 +42,17 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
+    int chargeTime_;
```
enochtsang: I don't believe this variable is being used and should be removed.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,55 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    {
+        totalAH_=BATTERY_AMP_HOUR_CAPACITY*initialStateOfChargePercent_/100;
```
enochtsang: This line I would recommend putting in the constructor since it does not require a `batteryData`.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,55 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+        isCharging_=true;
+        deltaAH_=-1*(BATTERY_AMP_HOUR_CAPACITY-totalAH_);
+
+    }
+    else{
+        isCharging_=false;
+        deltaAH_=totalAH_;
+    }
+    deltaAH_ /= batteryData.current;
+    return_time=zero_time.addMSecs(deltaAH_*HOUR_TO_SEC);
```
enochtsang: I would recommend removing line 81 and using 
`return_time = QTime(0, 0).addMSecs(deltaAH_*HOUR_TO_SEC);`

This way, there is less code and doesn't drop any readability.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,55 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+    if (previousTime_.isNull())
+    {
+        totalAH_=BATTERY_AMP_HOUR_CAPACITY*initialStateOfChargePercent_/100;
+        previousTime_=batteryData.time;
+    }
+    zero_time.setHMS(0,0,0,0);
+    deltaTime_=previousTime_.msecsTo(batteryData.time)/HOUR_TO_SEC;
+    deltaCurrent_=(batteryData.current+previousCurrent_)/2;
```
enochtsang: Here you're taking the average of the currents, renaming the variable to `avgCurrent` would be better.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,55 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    amph_=deltaTime_*deltaCurrent_;
+    totalAH_+=amph_;
```
enochtsang: The `amph_` is only used once, it would be fine to remove to make for shorter code. I would recommend just changing line 84 and 85 to:
`totalAH_ = deltaTime_ * deltaCurrent_;`. 
If you think the intermediate step is necessary to show what you are calculating, then `amph_` would only need to be a local variable and not a private member. Also, not abbreviating the name would better, so `amph_ --> ampHours_`

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,55 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+    if (batteryData.current < 0){
+        isCharging_=true;
```
enochtsang: Leaving this functionality to the designated function would be clearer. So instead of 
```
{
    return isCharging_;
}
```
We can remove a few lines of code without sacrificing readability with:
```
{
    return previousCurrent_ < 0;
}
```

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,55 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    return isCharging_;
 }
 
+
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    QTime null_time(0,0,0,0);
+    if (return_time.isNull())
+        return null_time;
```
enochtsang: Instead of having a local variable for this, I think it would be just as clear and shorter to replace line 67 and 69 with:
`return QTime(0, 0);`

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,55 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    if (batteryData.current < 0){
+        isCharging_=true;
+        deltaAH_=-1*(BATTERY_AMP_HOUR_CAPACITY-totalAH_);
+
+    }
+    else{
+        isCharging_=false;
+        deltaAH_=totalAH_;
+    }
+    deltaAH_ /= batteryData.current;
```
enochtsang: Keep this `deltaAH_` private member I'd say is good, but it's a little confusing that you're calculating it in a few different places. You are already calculating it in to `amph_` on line 84, so I'd suggest just using that calculation instead. Let me know if I'm misunderstanding something.

src/BatteryStateOfChargeService.cpp
```diff
@@ -26,15 +26,21 @@
  */
 
 #include "BatteryStateOfChargeService.h"
+#include <IOSTREAM>
```
enochtsang: This include breaks the compilation. Should just be `#include <iostream>`

src/BatteryStateOfChargeService.cpp
```diff
@@ -26,15 +26,21 @@
...
+#include <QTextStream>
+#include <QDebug>
+using std::cout; using std::endl;
```
enochtsang: This can just be 
```c++
using namespace std;
```

But overall, the `using` directive is frowned upon on our team. Explicitly stating `std::endl` is easier to understand.

src/BatteryStateOfChargeService.h
```diff
@@ -42,8 +42,16 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
+    QTime previousTime_;
+    QTime return_time;
+
+    double avgCurrent_;
+    double previousCurrent_;
+
+    double deltaTime_;
+
+    double totalAH_;
+    double deltaAH_;
```
enochtsang: This variable can just be a local variable in `addData`, doesn't need to be a private member. In general, you want to give a variable as small a scope as possible.

src/LogFileReader.cpp
```diff
@@ -74,12 +80,16 @@ bool LogFileReader::readAll(const QString& fileName)
...
-
-    // This is here to the compiler happy. Otherwise the compile
-    // will have an error warning about an unused variable. Remove this
-    // when you use it.
-    Q_UNUSED(line);
-    Q_UNUSED(batteryData);
-    return true;
+    QTime batteryDataTime = QTime::fromString(line.section(",",0,0), TIME_FORMAT);
+    double batteryDataVoltage= line.section(", ",1,1).toDouble();
+    double batteryDataCurrent = line.section(", ",2,2).toDouble();
```
enochtsang: The `", "` delimiter here deserves to be in a namespace on its own.

enochtsang:
Getting pretty close, there's still a comment near the top don't forgot about that one :)
Also, `batData.csv` and `batDataFailure.csv` are included twice in this pull request, only keep one of each.

src/LogFileReader.cpp
```diff
@@ -74,12 +80,16 @@ bool LogFileReader::readAll(const QString& fileName)
...
-    // This is here to the compiler happy. Otherwise the compile
-    // will have an error warning about an unused variable. Remove this
-    // when you use it.
-    Q_UNUSED(line);
-    Q_UNUSED(batteryData);
-    return true;
+    QTime batteryDataTime = QTime::fromString(line.section(SPACE_DELIMINTER,0,0), TIME_FORMAT);
+    double batteryDataVoltage= line.section(SPACE_DELIMINTER,1,1).toDouble();
+    double batteryDataCurrent = line.section(SPACE_DELIMINTER,2,2).toDouble();
+    if (batteryDataTime.isNull() || !batteryDataVoltage || !batteryDataCurrent){
```
enochtsang: To check for validity of the QTime, you should use `isValid()` rather than `isNull()` because you could get an invalid time that is not null.

enochtsang:
When running `batDataFailure.csv` 

The line
```
07:00:50.693, 3.553587, 7.953459, 6.52
```
gets parsed, but it should be an error since there are too many sections.

enochtsang:
Looks good, we'll just talk about some quick coding style. So make sure all your statement like `if` statements and `else` statements have brackets around them even if they're one line.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +46,54 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
 
 double BatteryStateOfChargeService::totalAmpHoursUsed() const
 {
-    return 0.0;
+    if (totalAmpHours_ > BATTERY_AMP_HOUR_CAPACITY)
+        return 0;
+    else if (totalAmpHours_ < 0)
+        return BATTERY_AMP_HOUR_CAPACITY;
+    else
+        return BATTERY_AMP_HOUR_CAPACITY - totalAmpHours_;
```
enochtsang: Missing brackets here

PR #10, opened 2015-11-21
Davis roman

DavisRomn:
Finished up and Fixed Errors


enochtsang:
So you have `batData.csv` and `batDataFailure.csv` in multiple places. If you could remove the unnecessary ones and just keep the ones you need to run the program it would make this pull request al lot  cleaner :)

src/BatteryStateOfChargeService.cpp
```diff
@@ -1,5 +1,8 @@
 #include "BatteryStateOfChargeService.h"
-
+#include <iostream>
+#include <QTime>
+#include <QDebug>
+using namespace std;
```
enochtsang: These 4 lines except for `#include <QTime>` are not being used and should be removed.

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +20,66 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    return QTime(0, 0, 0, 0).addMSecs(timeRemaining);
 }
 
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
+
+    QTime currentTime = batteryData.time;
+    changeTime = abs(initialTime.msecsTo(currentTime));
+    changeTime = changeTime * 2.77778e-7;
```
enochtsang: For string and numeric constants, it is a good programming practice to extract them out into named constants. In c++ one way of doing it is to use anonymous namespaces near the top of the file. The syntax would be like:

```c++
namespace
{
   const QString SOME_CONVERSION_FACTOR = 2.77778e-7;
}
```

The benefit for doing it this way is that if the format ever changes then you don't need to change it in multiple places.

src/BatteryStateOfChargeService.h
```diff
@@ -17,6 +17,25 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
 private:
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
-
+    double objectCurrent;
+    double inCurrent=0;
```
enochtsang: So when you want to initialize values, you should do it in the constructor. Having them initialized here is unclear and also doesn't compile on some compilers! On the compiler that comes with Qt, you get a warning.

src/BatteryStateOfChargeService.h
```diff
@@ -17,6 +17,25 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
+    double changeTime;
+    double inVoltage;
+    double Voltage;
+    QTime initialTime;
+    QTime currentTime;
+    double AmpHours;
+    double Current;
+    int counter;
+    double SumCurrent=0;
```
enochtsang: Same thing about putting this initialization in the constructor.

src/I_BatteryStateOfChargeService.h
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
 #include <QTime>
-struct BatteryData;
+#include <BatteryData.h>
```
enochtsang: So the `struct BatteryData` here was fine, this is called forward declaring it, don't worry about what that means for now. But having the full include `#include <BatteryData.h>` doesn't need to be in this file. You should have the full include in the cpp file for BatteryStateOfChargeService.

src/BatteryStateOfChargeService.h
```diff
@@ -17,6 +17,25 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
+    double TimeLeft;
+    double AverageCurrent;
+    double Hours;
+    double Minutes;
+    double Seconds;
+    double Milliseconds;
+    QTime t;
+    double AmpChange;
+    double dTime;
```
enochtsang: So you have a fair amount of variables which are not being used, you should remove them to keep this file as brief as possible.

src/LogFileReader.cpp
```diff
@@ -48,11 +48,27 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    QString timeString = sections.at(0);
+    //Converts timestring to time for bat data
+    batteryData.time = QTime::fromString(timeString, "hh:mm:ss.zzz");
+    //Takes V
+    bool voltageOK;
+    batteryData.voltage = sections.at(1) .toDouble(&voltageOK);
+    //Takes current
+    bool currentOK;
+    batteryData.current = sections.at(2) .toDouble(&currentOK);
+    if (!voltageOK || !currentOK || !batteryData.time.isValid())
```
enochtsang: To use the minimal amount of operators, I would recommend changing this if statement to this.

```
if(voltageOk && currentOk && batteryData.time.isValid())
{
    return true;
}
else
{ 
    return false;
}

File: src/BatteryStateOfChargeService.cpp [hunk 8]
```diff
@@ -16,21 +19,63 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+
+    QTime currentTime = batteryData.time;
+    changeTime = abs(initialTime.msecsTo(currentTime));
+    changeTime = changeTime * CONVERT_TO_MILLISECONDS;
+    initialTime = currentTime;
+    if ((Current > 0 && Current < 10000000000) || Current < 0 )
```
enochtsang: What's the reasoning behind this if statement? I'm not really understanding where these boundaries are coming from.

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +19,63 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    double AmpChange = (objectCurrent * changeTime);
+
+    AmpHours += AmpChange;
+
+    int counter;
+    counter++;
+    SumCurrent += inCurrent;
+    AverageCurrent = SumCurrent/counter;
+    double TimeLeft= (BATTERY_AMP_HOUR_CAPACITY-AmpHours)/AverageCurrent;
+    TimeLeft = qAbs(TimeLeft) * CONVERT_TO_MILLISECONDS;
```
enochtsang: This logic to calculate the time left should be done in that function, keeping code in their proper places makes it easier to read :)

src/BatteryStateOfChargeService.h
```diff
@@ -17,6 +17,15 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
 private:
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
-
+    double objectCurrent;
+    double inCurrent;
+    double changeTime;
+    QTime initialTime;
```
enochtsang: This variable is being updated with every new `initialTime` so the name is not really accurate, "initial" implies that this variable is for the very very first time.

src/BatteryStateOfChargeService.h
```diff
@@ -17,6 +17,15 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
 private:
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
-
+    double objectCurrent;
```
enochtsang: This could use a better name, I don't really understand what it's being used for right now.

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +19,63 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    {
+        objectCurrent = 0;
+    }
+    else
+    {
+        objectCurrent = (inCurrent + Current)/2;
+    }
+
+    double inVoltage = batteryData.voltage;
+    double Voltage = batteryData.voltage;
```
enochtsang: You're not really doing anything with these two lines right now, I don't think their necessary and can be removed

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +19,61 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+
+    QTime currentTime = batteryData.time;
+    changeTime = abs(prevTime.msecsTo(currentTime));
+    changeTime = changeTime * CONVERT_TO_MILLISECONDS;
+    prevTime = currentTime;
+    if (Current > 0 || Current < 0 )
```
enochtsang: Still not quite understanding what the purpose of this is because it's basically checking if `Current` is equal to 0.

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +19,61 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    Current = batteryData.current;
+
+    if (inCurrent == 0)
+    {
+        AvgCurrent = 0;
+    }
+    else
+    {
+        AvgCurrent = (inCurrent + Current)/2;
+    }
```
enochtsang: This entire section from lines 51-64 is fairly unclear. I would recommend understanding what the goal is in this section of code then rewriting it to be concise. Also, I would rename `inCurrent` since the `in` is not very descriptive.

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +19,61 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+    double inVoltage = batteryData.voltage;
+    double Voltage = batteryData.voltage;
```
enochtsang: The voltage here is not really used for anything, you can delete these two lines.

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +19,61 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    double AmpChange = (AvgCurrent * changeTime);
+
+    AmpHours += AmpChange;
+
+    int counter;
+    counter++;
```
enochtsang: When you initialize `int counter;` here, counter becomes a garbage value. So the `counter++` doesn't give you a value that you know what it is either, when you use it in `SumCurrent/counter` later, counter is a random value that you don't know what it is. Counter should be initialized in the constructor as a private member if you want to carry the value between function calls.

src/BatteryStateOfChargeService.h
```diff
@@ -17,6 +17,15 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
     const double initialStateOfChargePercent_;
-
+    double AvgCurrent;
+    double inCurrent;
+    double changeTime;
+    QTime prevTime;
+    QTime currentTime;
+    double AmpHours;
+    double Current;
+    double AverageCurrent;
```
enochtsang: What's the difference between `AverageCurrent` and `AvgCurrent`? Regardless, if they serve different purposes they should have different names.

src/BatteryStateOfChargeService.h
```diff
@@ -17,6 +17,14 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
-
+    double AvgCurrent;
+    double initialCurrent;
+    double changeTime;
+    QTime prevTime;
+    QTime currentTime;
+    double AmpHours;
+    double Current;
+    double dTime;
+    double SumCurrent;
```
enochtsang: This variable isn't being used anymore, should be deleted.

src/BatteryStateOfChargeService.h
```diff
@@ -17,6 +17,14 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
-
+    double AvgCurrent;
+    double initialCurrent;
+    double changeTime;
+    QTime prevTime;
+    QTime currentTime;
+    double AmpHours;
+    double Current;
```
enochtsang: You label the current time with `current` which is good, descriptive and I understand right away. This variable `Current` could use a prefix too, I wouldn't suggest `current` like you used for  `currentTime` (for obvious reasons).

Likewise, because you used `prev` for `prevTime`, following the same naming for `initialCurrent` would be better.

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +19,42 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    prevTime = currentTime;
+    initialCurrent = Current;
+    Current = batteryData.current;
+    AvgCurrent = (initialCurrent + Current)/2;
+    double AmpChange = (AvgCurrent * changeTime);
+
+    AmpHours += AmpChange;
+
+    dTime = prevTime.msecsTo(currentTime);
+    dTime = dTime * CONVERT_TO_MILLISECONDS;
```
enochtsang: Very well rewritten, much clearer!

enochtsang:
Stage 2: Code Structure Finished
Stage 3: Bugs and Glitches

enochtsang:
Stage 3: Bugs and Glitches. Finished
Stage 4: Coding Style

src/BatteryStateDisplayService.cpp
```diff
@@ -24,10 +24,20 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
...
     QTextStream(stdout) << "Voltage: " << batteryData.voltage
-        << " Current: " << batteryData.current
-        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed() << endl;
-
+                        << " Current: " << batteryData.current
+                        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed()<<endl;
     // TODO Print out time till it is depleted or charged.
+    if (batteryStateOfChargeService_.isCharging())
+    {
+        QTextStream(stdout)<<"Time left till full charge: "<<batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString("hh:mm:ss.zzz")<<endl;
```
enochtsang: This string constant should be abstracted out to a constant. Using the namespace like you did in other areas. Otherwise, the `<<` are considered operators and so should have spaces around them.

src/BatteryStateDisplayService.cpp
```diff
@@ -24,10 +24,20 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
...
+    }
+    else
+    {
+        QTextStream(stdout)<<"TIme left till fully discharged: "
```
enochtsang: Spaces around operators.

src/BatteryStateOfChargeService.cpp
```diff
@@ -1,13 +1,16 @@
 #include "BatteryStateOfChargeService.h"
-
+#include <QTime>
+#include <BatteryData.h>
 namespace
```
enochtsang: Includes should have a space separating them from the rest of the code, so:

```
#include <BatteryData.h>

namespace
```

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +19,44 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    {
+        return true;
+    }
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    double timeRemaining = (BATTERY_AMP_HOUR_CAPACITY-AmpHours)/prevCurrent;
+    timeRemaining = qAbs(timeRemaining)/CONVERT_TO_MILLISECONDS;
```
enochtsang: Spaces around operators in this section.

src/BatteryStateOfChargeService.h
```diff
@@ -17,6 +17,13 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
     const double initialStateOfChargePercent_;
-
+    double AvgCurrent;
+    double prevCurrent;
+    double changeTime;
+    QTime prevTime;
+    QTime currentTime;
+    double AmpHours;
+    double existingCurrent;
+    double dTime;
```
enochtsang: The convention is to have an underscore after private members like they're listed here.

enochtsang: In general, avoid abbreviations, this d is a good example of where not to abbreviate, write the full `delta` for this variable.

src/BatteryStateOfChargeService.h
```diff
@@ -17,6 +17,13 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
 private:
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
-
+    double AvgCurrent;
```
enochtsang: Variables should be camelCase, so new words are uppercase except for the first word. As a note, abbreviations should also be lowercase, like `theMvpOfTheNight`.

src/BatteryStateOfChargeService.h
```diff
@@ -17,6 +17,13 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
+    double prevCurrent;
+    double changeTime;
+    QTime prevTime;
```
enochtsang: This abbreviations is fine since it is commonly used. `prev`

src/LogFileReader.cpp
```diff
@@ -48,11 +48,27 @@ bool LogFileReader::readAll(const QString& fileName)
 bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) const
 {
     // TODO implement this first
+    QStringList sections = line.split(",");
+    if (sections.length() !=3)
```
enochtsang: Spaces around operators.

enochtsang:
Looks good!

PR #11, opened 2015-11-21
P wang

pwang8:


src/LogFileReader.cpp
```diff
@@ -47,12 +49,44 @@ bool LogFileReader::readAll(const QString& fileName)
...
+            if(allowedChars[count1] == line.toStdString().substr(count,1))
+            {
+                tempBool = true;
+            }
+        }
+        if(tempBool == false)
+        {
+            return false;
+        }
+    }
```
enochtsang: Instead of doing the error checking yourself, there are built in functions for what you are trying to do. You could split the string by it's delimiters, then for voltage and current take a look at http://doc.qt.io/qt-4.8/qstring.html#toDouble.

src/LogFileReader.cpp
```diff
@@ -47,12 +49,44 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    if(timeSplit.length()!=3)
+        {return false;}
+    QStringList secondSplit = timeSplit[2].split(".");
+    if(secondSplit.length()!=2)
+        {return false;}
+
+    QTime time = QTime(timeSplit[0].toInt(),    //hour
+                        timeSplit[1].toInt(),   //minute
+                        secondSplit[0].toInt(), //second
+                        secondSplit[1].toInt());//millisec
```
enochtsang: Instead of building the time manually, I would recommend using the built in function to convert a string into a QTime http://doc.qt.io/qt-5/qtime.html#fromString-1

src/test.cpp
```diff
@@ -0,0 +1,7 @@
+#include "test.h"
+
+test::test()
+{
+
+}
+
```
enochtsang: This and `test.h` are extra file and should be removed.

src/LogFileReader.cpp
```diff
@@ -2,8 +2,10 @@
 #include <QFile>
 #include <QString>
 #include <QTextStream>
+#include <QTime>
 
 #include "LogFileReader.h"
+#include <iostream>
```
enochtsang: You're not using this include and so it should be removed.

src/LogFileReader.cpp
```diff
@@ -47,12 +49,44 @@ bool LogFileReader::readAll(const QString& fileName)
 // File input is a csv file in the format of hh:mm:ss:zzz, voltage, current
 bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) const
 {
-    // TODO implement this first
+    unsigned int allowedCharLength = 15;
+    std::string allowedChars [allowedCharLength] = {"1","2","3","4","5","6","7","8","9","0",":",",",".","-"," "};
```
enochtsang: When you can, you should try to use the built in types, in this case ` QString` would fit better instead of `std::string`.

src/LogFileReader.cpp
```diff
@@ -2,8 +2,10 @@
 #include <QFile>
 #include <QString>
 #include <QTextStream>
+#include <QTime>
```
enochtsang: Because this was included in the header file, this include is not necessary.

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +20,57 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+    if(myTime.isNull())
+    {
+        prevTime = new QTime(0,0,0,0);
+    }
+    myVoltage = batteryData.voltage;
+    prevCurrent = myCurrent;
+    myCurrent = batteryData.current;
+    prevTime = new QTime(myTime.hour(),myTime.minute(),myTime.second(),myTime.msec());
```
enochtsang: Everytime you use `new` you are allocating dynamic memory, which also means you are in charge of deallocating the same memory using `delete`. In this case, you have a memory leak because you are never deallocating this memory.

src/BatteryStateOfChargeService.h
```diff
@@ -17,6 +18,11 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
 private:
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
+    QTime myTime;
+    QTime *prevTime;
+    double myVoltage;
```
enochtsang: I don't believe storing voltage is necessary in this class, so this private member is not needed.

src/BatteryStateOfChargeService.h
```diff
@@ -17,6 +18,11 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
+    double myCurrent;
```
enochtsang: Using `my` is not very descriptive to indicate for the current `batteryData`, you should look for another prefix to indicate that :)

src/BatteryStateOfChargeService.cpp
```diff
@@ -1,13 +1,17 @@
 #include "BatteryStateOfChargeService.h"
+#include <stdio.h>
 
 namespace
 {
     const double BATTERY_AMP_HOUR_CAPACITY = 123.0;
+    double ampHourTotal;
+    double ampHourUsed;
```
enochtsang: Usually we reserve this area for constants, if you need a variable held between calls in your class, you should use a private member.

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +20,57 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    if(isCharging())
+    {
+        ampHoursLeft = BATTERY_AMP_HOUR_CAPACITY - ampHourTotal;
+    }
+    else
+    {
+        ampHoursLeft = ampHourTotal;
+    }
+    int current = myCurrent;
+    if(myCurrent < 0)
```
enochtsang: You've already checked this condition previously with `isCharging()`, you should put the logic in these statements into where you've already checked the condition. So:
```c++
    double ampHoursLeft;
    int current = myCurrent;
    if(isCharging())
    {
        ampHoursLeft = BATTERY_AMP_HOUR_CAPACITY - ampHourTotal;
    }
    else
    {
        ampHoursLeft = ampHourTotal;
        current = -myCurrent;
    }
```

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +20,57 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    {
+        current *= -1;
+    }
+    double msecLeft = (ampHoursLeft * 3600000) / current;
```
enochtsang: This `3600000` is what we refer to as a "magic number" and should be abstracted into a constant in an anonymous namespace. We usually do that by adding to the top:
```
namespace 
{
    int CONVERSION_FACTOR = 3600000;
}
```
Of course, pick a better name than `CONVERSION_FACTOR` :)

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +20,57 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    QTime QTimeLeft = QTime(0,0,0,0);
+    QTimeLeft = QTimeLeft.addMSecs(msecLeft);
```
enochtsang: These two lines can be compressed into one line using, `QTime timeLeft = QTime(0, 0).addMsecs(msecLeft);`

src/BatteryStateOfChargeService.cpp
JordanHeinrichs: Just to further expand on why, if there were multiple instances of BatteryStateOfChargeServices within the program then all of the instances would share the same ampHourUsed and ampHourTotal variable therefore would not work.

src/BatteryStateOfChargeService.cpp
JordanHeinrichs: Make sure it to be const in front, i.e.
```c++
namespace 
{
    const int CONVERSION_FACTOR = 3600000;
}
```

src/BatteryStateOfChargeService.cpp
pwang8: Oh i see, so namespace is a static field then. I should've done more research on what a namespace is, had put them there "because it worked" and because I had trouble declaring it as private. I'll keep that in mind thanks! :)

enochtsang: Just to clarify, the namespace isn't really a part of the class (as a static field or otherwise), it's something that can be used in that `.cpp` file.

src/BatteryStateOfChargeService.cpp
JordanHeinrichs: While in this case I don't think you need to use dynamic memory for prevTime but if you decide to use it I recommend using http://doc.qt.io/qt-5/qscopedpointer.html. It takes care of the memory management for you.

src/LogFileReader.cpp
```diff
@@ -47,12 +47,25 @@ bool LogFileReader::readAll(const QString& fileName)
...
 bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) const
 {
-    // TODO implement this first
+    //example input= 07:01:07.024, 3.538654, 25.523571
+    bool convertedOkay;
+    QStringList lineSplit = line.split(",");
+    if(lineSplit.length()!= 3)
+        {return false;}
+
+    QTime time = QTime::fromString(lineSplit[0],"hh:mm:ss.zzz");
```
enochtsang: You should move this `"hh.:mm:ss.zzz"` and `","` constant into an anonymous namespace constant variable at the top.

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +19,53 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    {
+        ampHoursLeft = BATTERY_AMP_HOUR_CAPACITY - ampHourTotal_;
+        current *= -1;
+    }
+    else
+    {
+        ampHoursLeft = ampHourTotal_;
+    }
+    double msecLeft = (ampHoursLeft * HOURS_TO_MSECS) / current;
+    QTime QTimeLeft = QTime(0,0,0,0).addMSecs(msecLeft);
```
enochtsang: It's not necessary to name variables with their type, doing this is called hungarian notation and generally that's not how we name things on the team and is frowned upon. `timeLeft` should suffice :)

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +19,53 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 {
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+    if(currTime_.isNull())
+    {
+        prevTime_ = QTime(0,0,0,0);
+    }
+    prevCurrent_ = currCurrent_;
+    currCurrent_ = batteryData.current;
+    prevTime_ = QTime(currTime_.hour(),currTime_.minute(),currTime_.second(),currTime_.msec());
```
enochtsang: You can just use the copy constructor here, `prevTime = currTime;` would suffice.

src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +19,53 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    currTime_ = batteryData.time;
+
+    //ampHourCalculation is put here instead of totalAmpHoursUsed() due to the method being a const
+    double diffOfTimeMSec = prevTime_.msecsTo(currTime_);
+    double diffOfTime = diffOfTimeMSec/HOURS_TO_MSECS;
```
enochtsang: Compressing these two lines into one variable would be fine I think. 
```c++
double diffOfTime = prevTime_.msecsTo(currTime_) / HOURS_TO_MSECS;
```

enochtsang:
Does it compile on your version? Mine gets this error
```
g++ -c -pipe -O2 -Wall -W -D_REENTRANT -fPIE -DQT_NO_DEBUG -DQT_CORE_LIB -I../../../../Qt5.0.2/5.0.2/gcc/mkspecs/linux-g++ -I. -I../../../../Qt5.0.2/5.0.2/gcc/include -I../../../../Qt5.0.2/5.0.2/gcc/include/QtCore -I../release/.moc -o ../release/.obj/BatteryStateOfChargeService.o BatteryStateOfChargeService.cpp
BatteryStateOfChargeService.cpp: In constructor ‘BatteryStateOfChargeService::BatteryStateOfChargeService(double)’:
BatteryStateOfChargeService.cpp:13:5: error: ‘ampHourTotal_’ was not declared in this scope
     ampHourTotal_ = (initialStateOfChargePercent_/100) * BATTERY_AMP_HOUR_CAPACITY;
     ^
BatteryStateOfChargeService.cpp: In member function ‘virtual double BatteryStateOfChargeService::totalAmpHoursUsed() const’:
BatteryStateOfChargeService.cpp:22:12: error: ‘ampHourUsed_’ was not declared in this scope
     return ampHourUsed_;
            ^
BatteryStateOfChargeService.cpp: In member function ‘virtual QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const’:
BatteryStateOfChargeService.cpp:40:52: error: ‘ampHourTotal_’ was not declared in this scope
         ampHoursLeft = BATTERY_AMP_HOUR_CAPACITY - ampHourTotal_;
                                                    ^
BatteryStateOfChargeService.cpp:45:24: error: ‘ampHourTotal_’ was not declared in this scope
         ampHoursLeft = ampHourTotal_;
                        ^
BatteryStateOfChargeService.cpp: In member function ‘virtual void BatteryStateOfChargeService::addData(const BatteryData&)’:
BatteryStateOfChargeService.cpp:69:5: error: ‘ampHourTotal_’ was not declared in this scope
     ampHourTotal_ += changeInAmpHour;
     ^
BatteryStateOfChargeService.cpp:70:5: error: ‘ampHourUsed_’ was not declared in this scope
     ampHourUsed_ = BATTERY_AMP_HOUR_CAPACITY- ampHourTotal_;
     ^
BatteryStateOfChargeService.cpp: In member function ‘virtual double BatteryStateOfChargeService::totalAmpHoursUsed() const’:
BatteryStateOfChargeService.cpp:23:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
make: *** [../release/.obj/BatteryStateOfChargeService.o] Error 1
```

src/LogFileReader.cpp
```diff
@@ -47,12 +51,25 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    QStringList lineSplit = line.split(",");
+    if(lineSplit.length()!= 3)
+        {return false;}
+
+    QTime time = QTime::fromString(lineSplit[0],TIME_FORMAT);
+    if(!time.isValid())
+        {return false;}
+
+    double voltage = lineSplit[1].toDouble(&convertedOkay);
+    if(!convertedOkay||voltage<0)
```
enochtsang: Is there any rationale for needing `voltage < 0`? This shouldn't be required

enochtsang:
So in general it looks good on coding style, just one major thing you're missing.

There should be spaces around operators, so like:
```c++
for(int i = 0; i < 5; i++)
``` 
instead of 
```c++
for(int i=0; i<5;i++);
```

src/LogFileReader.cpp
```diff
@@ -47,12 +51,25 @@ bool LogFileReader::readAll(const QString& fileName)
...
+        {return false;}
+
+    double current = lineSplit[2].toDouble(&convertedOkay);
+    if(!convertedOkay)
+        {return false;}
```
enochtsang: These `if` statements I would give two ways of writing them:
```c++
if(!convertedOkay) { return false; }
```
Note the spaces around the `{` brackets.

The other way that we see more regularly is:
```c++
if(!convertedOkay)
{
    return false;
}
```

enochtsang:
Looks good! Finished!

PR #12, opened 2016-04-09
initial commit

ikazuka36:
create pull request


src/BatteryStateDisplayService.cpp
```diff
@@ -51,7 +51,8 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
 void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& batteryData)
 {
     batteryStateOfChargeService_.addData(batteryData);
-
+    batteryStateOfChargeService_.isCharging();
+    qDebug()<<"Time finish charging: " <<batteryStateOfChargeService_.timeWhenChargedOrDepleted();
     QTextStream(stdout) << "Voltage: " << batteryData.voltage
```
enochtsang: Since this section of code here is already using `QTextStream` you should follow suit and use the same for debugging rather than using `QDebug`.

ikazuka36: I don't think QTextStream is able to print out QTIme. I got error display when I tried that

enochtsang: So at that, you would want to convert the QTime into a QString using the `toString()` function. I think that's the name of the function, but the docs are down so I'm not sure ;o

src/LogFileReader.cpp
```diff
@@ -75,11 +78,37 @@ bool LogFileReader::readAll(const QString& fileName)
...
     // TODO implement this first
+    QStringList lines = line.split(',');
+    bool ok;
+    if(lines.count()!=3)
+        return false;
+    QStringList lines_1 = lines[0].split(':');
+    QStringList lines_2 = lines[1].split('.');
+    int value = lines_1[0].toInt(&ok);
+    if((!ok)||(value>24)||(value<0))
+        return false;
```
enochtsang: Instead of parsing the time yourself, I'd recommend using QTime's `fromString` function. Here is a link to the documentation

http://doc.qt.io/qt-5/qtime.html#fromString-1

You can use the `isValid()` function afterwards to check if the conversion was successful.

src/LogFileReader.cpp
```diff
@@ -75,11 +78,37 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    }
+    QTime t = QTime(lines_1[0].toInt(),lines_1[1].toInt(),lines_1[2].toDouble());
+    batteryData.time = t;
+    batteryData.voltage = lines[1].toDouble();
+    batteryData.current = lines[2].toDouble();
     // This is here to the compiler happy. Otherwise the compile
     // will have an error warning about an unused variable. Remove this
     // when you use it.
     Q_UNUSED(line);
     Q_UNUSED(batteryData);
```
enochtsang: These line should be removed now, they were there previously to avoid compiler warnings.

src/BatteryStateOfChargeService.cpp
```diff
@@ -35,29 +36,49 @@ namespace
...
-    return 0.0;
+    return totalcurrent;
 }
 
 bool BatteryStateOfChargeService::isCharging() const
 {
-    return false;
+    if (newcurrent<=0)
+        qDebug()<<"is charging";
+    else
```
enochtsang: Instead of the qDebug here, you should probably be returning true.

src/BatteryStateOfChargeService.h
```diff
@@ -44,6 +44,14 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
-
+    double oldcurrent;
+    double newcurrent;
+    double totalcurrent;
+    QTime newtime;
+    QTime oldtime;
+    QTime et;
+    int timeused;
+    double s1;
+    double s2;
```
enochtsang: A lot of these private members are only being used in the `addData` function. We want to keep the scope of variables as small as we can, so variables like s1, s2, should just be local variables inside the `addData` function.

src/BatteryStateOfChargeService.h
```diff
@@ -44,6 +44,14 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
 private:
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
-
+    double oldcurrent;
+    double newcurrent;
+    double totalcurrent;
+    QTime newtime;
+    QTime oldtime;
+    QTime et;
```
enochtsang: `et` could use a better name, without context it's hard to tell what it's supposed to represent! Long names beat cryptic short names everyday :)

src/BatteryStateOfChargeService.cpp
```diff
@@ -35,29 +36,49 @@ namespace
...
-{
+{   
+    newcurrent = batteryData.current;
+    newtime = batteryData.time;
+    s1 = QTime(0,0,0).secsTo(newtime);
+    s2 = QTime(0,0,0).secsTo(oldtime);
+    timeused = s1-s2;
+    totalcurrent = newcurrent - oldcurrent;
+    double avg = (fabs(newcurrent)+fabs(oldcurrent)) *3600 / timeused;
+    int tneed =(BATTERY_AMP_HOUR_CAPACITY - newcurrent)*3600/avg;
```
enochtsang: You're using `3600` quite often here and makes it a little difficult to tell what's actually going on in those lines. What would be better is making it a constant. We usually keep our constants in the `.cpp` file in an anonymous namespace at the top.

```c++
namespace 
{
    const int SOME_CONVERSION_FACTOR = 3600;
}
```

enochtsang:
Overall not bad! Some style things you should fix:

- Always use braces around `if` and `else` statements including one liner if and else statements. It makes it easier to add code in later. It also keeps the code slightly easier to read since it separates blocks of logic.
- Watch the white spacing, there are some places you are missing tabs and spaces. After `#include` directives there should be a space, and all binary operators like `+`, `-`, etc should have spaces around them. 
- We usually use trailing underscores after private members so it's immediately clear what is a private member what is a local variable. like `int oldCurrent_`.
- Class members, local variables, and arguments all should be `CamelCase` to be consistent with the rest of the code, in terms of this pull request that's not the only one that's off. But you should also note types like classes or enums are `CapCase`, constants are `ALL_CAPS`.

src/BatteryStateOfChargeService.cpp
```diff
@@ -35,29 +36,49 @@ namespace
...
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
-{
+{   
+    newcurrent = batteryData.current;
+    newtime = batteryData.time;
+    s1 = QTime(0,0,0).secsTo(newtime);
+    s2 = QTime(0,0,0).secsTo(oldtime);
+    timeused = s1-s2;
+    totalcurrent = newcurrent - oldcurrent;
+    double avg = (fabs(newcurrent)+fabs(oldcurrent)) *3600 / timeused;
```
bill-luu: The build fails because _fabs_ was not declared in this scope, and I can't find it anywhere else in your program. It looks like you forgot to define it.

src/LogFileReader.cpp
```diff
@@ -75,11 +78,37 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    bool ok;
+    if(lines.count()!=3)
+        return false;
+    QStringList lines_1 = lines[0].split(':');
+    QStringList lines_2 = lines[1].split('.');
+    int value = lines_1[0].toInt(&ok);
+    if((!ok)||(value>24)||(value<0))
+        return false;
 
+    int m = lines_1[1].toInt(&ok);
```
bill-luu: Just going to reiterate what Enoch said about variable names, make them nice and descriptive!

src/BatteryStateOfChargeService.cpp
ikazuka36: fabs is a function in the <math.h> used to get the absolute value of the parameter so we don't have to declare in the scope.

src/LogFileReader.cpp
ikazuka36: I'm deleting the whole things and make new lines following Enoch suggestion but yeah gotta stop making cryptic variable names :D

src/BatteryStateOfChargeService.cpp
```diff
@@ -24,40 +24,65 @@
...
+{   
+    newcurrent_ = batteryData.current;
+    double oldcurrent = 0;
+    QTime newtime = batteryData.time;
+    QTime oldtime = newtime;
+    double s1 = QTime(0,0,0).secsTo(newtime);
+    double s2 = QTime(0,0,0).secsTo(oldtime);
+    int timeused = s1 - s2;
+    totalcurrent_ = newcurrent_ - oldcurrent;
+    double avg = (fabs(newcurrent_)+fabs(oldcurrent)) * SH_CONVERT / timeused;
```
bill-luu: I assume it worked Ok on yours, but I had to add an
`#include <math.h>`
in order for `fabs` to be recognized, it may just be a configuration thing.

src/BatteryStateOfChargeService.cpp
```diff
@@ -31,6 +31,9 @@
 namespace
 {
     const double BATTERY_AMP_HOUR_CAPACITY = 123.0;
+    const int SH_CONVERT = 3600;
```
bill-luu: You might as well expand `SH` to something like `SECONDS_TO_HOURS`

src/BatteryStateOfChargeService.cpp
```diff
@@ -31,6 +31,9 @@
...
+    const int M_CONVERT = 60;
```
bill-luu: Same as above, just do `MINUTE_CONVERT` or something along those lines.

src/BatteryStateOfChargeService.cpp
```diff
@@ -46,39 +49,40 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    newcurrent_ = batteryData.current;
+    double oldcurrent = 0;
+    QTime newtime = batteryData.time;
+    QTime oldtime = newtime;
+    double s1 = QTime(0,0,0).secsTo(newtime);
+    double s2 = QTime(0,0,0).secsTo(oldtime);
+    int timeused = s1 - s2;
+    totalcurrent_ = newcurrent_ - oldcurrent;
+    double avg = (fabs(newcurrent_)+fabs(oldcurrent)) * SH_CONVERT / timeused;
+    int tneed =(BATTERY_AMP_HOUR_CAPACITY - newcurrent_)* SH_CONVERT / avg;
```
bill-luu: `tneed` can be expanded to something like `timeNeeded`

src/LogFileReader.cpp
```diff
@@ -75,11 +80,25 @@ bool LogFileReader::readAll(const QString& fileName)
 bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) const
 {
     // TODO implement this first
-
+    QStringList lines = line.split(',');
```
bill-luu: `,` is what we can call a delimiter, and we can put it in the namespace as a const QString.

src/LogFileReader.cpp
```diff
@@ -75,11 +80,25 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    bool ok;
+    //I have to split up the time and round up the second into an integer in order for fromString to work
+    //Not sure how to use fromString with 3 decimals in second
+    QStringList timelist = lines[0].split(':');
+    double second = timelist[2].toDouble(&ok);
+    second = round(second);
+    timelist[2] = QString::number(second);
+    lines[0]= timelist.join(":");
+    QTime time = QTime::fromString(lines[0],"hh:mm:s");
```
bill-luu: Similarly to the delimiter, `"hh:mm:s"` can be put into the namespace as a const QString as well. 

Also, the format of the times in batData.csv is actually `hh:mm:ss.zzz`.

src/BatteryStateDisplayService.cpp
```diff
@@ -51,10 +51,11 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
 void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& batteryData)
 {
     batteryStateOfChargeService_.addData(batteryData);
-
+    batteryStateOfChargeService_.isCharging();
     QTextStream(stdout) << "Voltage: " << batteryData.voltage
         << " Current: " << batteryData.current
-        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed() << endl;
+        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed()
+        <<"Time finish charging: " <<batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString()<< endl;
```
benji373: We should have grammatically sentences.  "Time finish charging:" should be "Time until finished charging:" or something similar.

src/BatteryStateOfChargeService.h
```diff
@@ -44,6 +44,8 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
 private:
     // Convention is to have an _ after private members
```
benji373: We could clean this up a bit by removing some of these comments

src/LogFileReader.cpp
```diff
@@ -75,11 +84,17 @@ bool LogFileReader::readAll(const QString& fileName)
...
-
+    QStringList lines = line.split(COMMA);
+    QTime time = QTime::fromString(lines[0],TIMEFORMAT);
+    if (time.isValid())
+    {
+        batteryData.time = time;
+    }
+    batteryData.voltage = lines[1].toDouble();
+    batteryData.current = lines[2].toDouble();
     // This is here to the compiler happy. Otherwise the compile
```
benji373: now that we've removed these lines for the unused variables, as above we can clean this up by removing the comments.

src/BatteryStateDisplayService.cpp
```diff
@@ -24,9 +24,9 @@
  *
  *  For further contact, email <software@calgarysolarcar.ca>
  */
-
+#include <QDebug>
 #include <QTextStream>
-
+#include <BatteryStateOfChargeService.h>
```
enochtsang: quote marks

src/BatteryStateDisplayService.cpp
```diff
@@ -51,10 +51,11 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
 void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& batteryData)
 {
     batteryStateOfChargeService_.addData(batteryData);
-
+    batteryStateOfChargeService_.isCharging();
     QTextStream(stdout) << "Voltage: " << batteryData.voltage
         << " Current: " << batteryData.current
-        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed() << endl;
+        << " Total Ah used:  " << batteryStateOfChargeService_.totalAmpHoursUsed()
+        <<" Time until fully charged: " <<batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString()<< endl;
```
enochtsang: missing spaces around operators

src/BatteryStateOfChargeService.cpp
```diff
@@ -24,40 +24,65 @@
  *
  *  For further contact, email <software@calgarysolarcar.ca>
  */
-
+#include<QDebug>
```
enochtsang: space here

src/BatteryStateOfChargeService.cpp
```diff
@@ -24,40 +24,65 @@
...
 #include "BatteryStateOfChargeService.h"
-
+#include <QTime>
+#include <math.h>
 namespace
 {
     const double BATTERY_AMP_HOUR_CAPACITY = 123.0;
+    const int SECONDS_TO_HOURS = 3600;
+    const int SECONDS_TO_MINUTUES = 60;
+
```
enochtsang: extra space here

src/BatteryStateOfChargeService.cpp
```diff
@@ -24,40 +24,65 @@
...
+    return totalcurrent_;
 }
 
 bool BatteryStateOfChargeService::isCharging() const
 {
-    return false;
+    if (newcurrent_ <= 0)
+    {   return true;}
+    else
+    {   return false;}
```
enochtsang: different lines

src/BatteryStateOfChargeService.h
```diff
@@ -42,8 +42,8 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     void addData(const BatteryData& batteryData);
 
 private:
-    // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
-
-    // Recommend adding necessary variables here
+    double newcurrent_;
+    double totalcurrent_;
+    QTime estimatetime_;
```
enochtsang: Follow naming conventions

src/BatteryStateOfChargeService.cpp
```diff
@@ -24,40 +24,65 @@
...
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    return estimatetime_;
```
enochtsang: `estimatedTime_`

src/BatteryStateOfChargeService.cpp
```diff
@@ -24,40 +24,65 @@
...
-    // Update your variables here.
+{   
+    newcurrent_ = batteryData.current;
+    double oldcurrent = 0;
+    QTime newtime = batteryData.time;
+    QTime oldtime = newtime;
+    double s1 = QTime(0,0,0).secsTo(newtime);
+    double s2 = QTime(0,0,0).secsTo(oldtime);
+    int timeused = s1 - s2;
+    totalcurrent_ = newcurrent_ - oldcurrent;
```
enochtsang: oldcurrent is always 0 here

src/BatteryStateOfChargeService.cpp
```diff
@@ -24,40 +24,65 @@
...
-{
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+{   
+    newcurrent_ = batteryData.current;
+    double oldcurrent = 0;
+    QTime newtime = batteryData.time;
+    QTime oldtime = newtime;
+    double s1 = QTime(0,0,0).secsTo(newtime);
+    double s2 = QTime(0,0,0).secsTo(oldtime);
```
enochtsang: spaces after commas

enochtsang: rename s1 s2

src/BatteryStateOfChargeService.cpp
```diff
@@ -24,40 +24,65 @@
...
+    int timeused = s1 - s2;
+    totalcurrent_ = newcurrent_ - oldcurrent;
+    double avg = (fabs(newcurrent_)+fabs(oldcurrent)) * SECONDS_TO_HOURS / timeused;
```
enochtsang: no fabs here

src/BatteryStateOfChargeService.cpp
```diff
@@ -24,40 +24,65 @@
...
+#include "BatteryData.h"
 #include "BatteryStateOfChargeService.h"
-
+#include <QTime>
+#include <math.h>
 namespace
 {
     const double BATTERY_AMP_HOUR_CAPACITY = 123.0;
+    const int SECONDS_TO_HOURS = 3600;
+    const int SECONDS_TO_MINUTUES = 60;
```
enochtsang: spelling error

src/LogFileReader.cpp
```diff
@@ -24,12 +24,21 @@
...
-
+#include <QString>
 #include <QDebug>
 #include <QFile>
 #include <QString>
 #include <QTextStream>
+#include <iostream>
+#include <math.h>
+#include <cstring>
+#include <string>
```
enochtsang: remove old inclues

src/LogFileReader.cpp
```diff
@@ -74,12 +83,14 @@ bool LogFileReader::readAll(const QString& fileName)
...
-    Q_UNUSED(batteryData);
+    QStringList lines = line.split(COMMA);
+    QTime time = QTime::fromString(lines[0],TIMEFORMAT);
+    if (time.isValid())
+    {
+        batteryData.time = time;
+    }
+    batteryData.voltage = lines[1].toDouble();
+    batteryData.current = lines[2].toDouble();
     return true;
```
enochtsang: Handle error checking

src/BatteryStateOfChargeService.cpp
ikazuka36: I actually update the oldCurrent at the end of the code rather than at the beginning.

enochtsang: The `oldcurrent` variable is still always 0 at this point regardless of when you update it. 
`totalcurrent_ = newcurrent_ - oldcurrent;` will always be `totalcurrent_ = new current_ - 0;` which isn't really all that useful :P

src/LogFileReader.cpp
```diff
@@ -74,12 +79,32 @@ bool LogFileReader::readAll(const QString& fileName)
...
-    // will have an error warning about an unused variable. Remove this
-    // when you use it.
-    Q_UNUSED(line);
-    Q_UNUSED(batteryData);
+    if (lines.count() !=    3)
+    {
+        return false;
+    }
+
+    for (int c = 1; c<=2; c++)
```
enochtsang: It's much more standard to use `i` in for loops. The `c` has potential to confuse people to look for context where there isnt' any.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,48 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 }
 
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
-{
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+{   
+
+    QTime newTime = batteryData.time;
+    QTime oldTime = newTime;
```
enochtsang: Here `oldTime` and `newTime` will be equal, you probably want to set the `oldTime` before you set `newTime`.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,48 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    double new_Time = QTime(0, 0, 0).secsTo(newTime);
+    double old_Time = QTime(0, 0, 0).secsTo(oldTime);
```
enochtsang: Instead of being ambiguous between `new_Time` and `newTime` it would be better to actually name the difference between the two. How about `newTimeSecs` instead?

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,48 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    int TimeUsed = new_Time - old_Time;
+
+    newCurrent_ = batteryData.current;
+    double oldCurrent = 0;
+    totalCurrent_ = newCurrent_ - oldCurrent;
+    double Average = (newCurrent_ + oldCurrent) * SECONDS_TO_HOURS / TimeUsed;
```
enochtsang: What average are you calculating? If you want the average of two currents, all you need is `(newCurrent_ + oldCurrent) / 2`

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,48 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+    int timeNeeded =(BATTERY_AMP_HOUR_CAPACITY - newCurrent_) * SECONDS_TO_HOURS / Average;
+    new_Time = new_Time + timeNeeded;
```
enochtsang: What's the purpose of adding `timeNeeded` here?

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,48 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+    int Hour = new_Time / SECONDS_TO_HOURS;
+    int Minute = (new_Time - (Hour * SECONDS_TO_HOURS)) / SECONDS_TO_MINUTES;
+    int Second = new_Time - (Hour * SECONDS_TO_HOURS) - (Minute * SECONDS_TO_MINUTES);
+    estimatedTime_ = QTime(Hour, Minute, Second);
```
enochtsang: Since you already have this time in milliseconds, it would be better to use `QTime(0, 0).addMSecs(int ms)` instead of finding the hours, minutes, and seconds yourself.

ikazuka36: I had it in second so I'd use addSecs instead

src/BatteryStateOfChargeService.cpp
ikazuka36: It's the time needed to fully charge the battery

src/BatteryStateOfChargeService.cpp
ikazuka36: Shouldn't it be (newCurrent_ + oldCurrent) / delta t ? And since most of the times are in seconds while other variable are per hour I have to convert it into per seconds to calculate the Average

src/BatteryStateOfChargeService.h
```diff
@@ -42,8 +42,11 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     void addData(const BatteryData& batteryData);
 
 private:
-    // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
+    double newCurrent_;
+    double totalCurrent_;
+    QTime estimatedTime_;
+    double oldCurrent_ = 0;
+    QTime oldTime_;
```
enochtsang: Intialize in constructor

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    else
+    {
+        return false;
+    }
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    return estimatedTime_;
```
enochtsang: more logic here

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
-{
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+{   
+
+    QTime newTime = batteryData.time;
+    double new_TimeDouble = QTime(0, 0, 0).secsTo(newTime);
+    double old_TimeDouble = QTime(0, 0, 0).secsTo(oldTime_);
```
enochtsang: rneame to msecs

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +48,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    int TimeUsed = new_TimeDouble - old_TimeDouble;
+
+    newCurrent_ = batteryData.current;
+    totalCurrent_ = newCurrent_ - oldCurrent_;
+    double Average = (newCurrent_ + oldCurrent_) * SECONDS_TO_HOURS / TimeUsed;
```
enochtsang: rename to `averageCurrent`, divide by two for average

src/BatteryStateDisplayService.cpp
```diff
@@ -51,10 +51,20 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
...
+    else
+    {
+        QTextStream(stdout) << "Voltage: " << batteryData.voltage
         << " Current: " << batteryData.current
-        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed() << endl;
+        << " Total Ah used:  " << batteryStateOfChargeService_.totalAmpHoursUsed()
+        << " Time until fully depleted: " << batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString() << endl;
 
-    // TODO Print out time till it is depleted or charged.
+    }
```
enochtsang: There is duplicated code here, you should move the duplicated code outside of the if else statement.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +49,39 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    return estimatedTime;
+}
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+    newTime_ = batteryData.time;
+    newCurrent_ = batteryData.current;
+
+    double TimeUsed = newTime_.msecsTo(oldTime_);
```
enochtsang: `TimeUsed` should start lowercase

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +49,39 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    double averageCurrent = (newCurrent_ + oldCurrent_) / 2;
+    totalAmpUsed_ = initialAmpUsed_ + averageCurrent  / (TimeUsed / SECONDS_TO_MILISECONDS);
+
+    oldTime_ = newTime_;
+    oldCurrent_ = newCurrent_;
```
enochtsang: These variables should be set at the beginning of this function. Because if any other functions use variables, they won't be very helpful, since the "old" and "new" variables will be the same at the end of this function.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +49,39 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+    newTime_ = batteryData.time;
+    newCurrent_ = batteryData.current;
+
+    double TimeUsed = newTime_.msecsTo(oldTime_);
+    double averageCurrent = (newCurrent_ + oldCurrent_) / 2;
+    totalAmpUsed_ = initialAmpUsed_ + averageCurrent  / (TimeUsed / SECONDS_TO_MILISECONDS);
```
enochtsang: You should be updating `totalAmpUsed_` based on `totalAmpUsed_`. By updating it according to `initialAmpUsed_` you will be having a number similar to the initial amps rather than an updating total amps.

src/BatteryStateOfChargeService.h
```diff
@@ -42,8 +42,13 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
-
-    // Recommend adding necessary variables here
+    double newCurrent_;
+    double totalCurrent_;
+    QTime estimatedTime_;
+    double oldCurrent_;
+    QTime oldTime_ = QTime(0,0,0);
+    QTime newTime_;
+    double totalAmpUsed_ = 0;
+    double initialAmpUsed_;
```
enochtsang: You have a couple private members here that are not being used anymore. 

You also are trying to initialize variables here, this is not valid C++ in some compilers and is also not good practice. It is better to initialize them in the constructor.

src/BatteryStateOfChargeService.h
```diff
@@ -42,8 +42,13 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
 private:
-    // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
-
-    // Recommend adding necessary variables here
+    double newCurrent_;
+    double totalCurrent_;
+    QTime estimatedTime_;
+    double oldCurrent_;
+    QTime oldTime_ = QTime(0,0,0);
```
enochtsang: Private members type custom objects (not primitive types) will often be initialied with their default constructor. In this case `oldTime_` would be constructed with `QTime()` which would effectively be the same as `QTime(0, 0, 0)`.

src/LogFileReader.cpp
```diff
@@ -67,19 +71,38 @@ bool LogFileReader::readAll(const QString& fileName)
...
-    // This is here to the compiler happy. Otherwise the compile
-    // will have an error warning about an unused variable. Remove this
-    // when you use it.
-    Q_UNUSED(line);
-    Q_UNUSED(batteryData);
+    for (int i = 1; i<=2; i++)
+    {
+        lines[i].toDouble(&ok);
+        if (!ok)
+            return false;
```
enochtsang: All if else statements should be enclosed in braces regardless if they're one liners or not. This is because code quickly becomes not one liners, and also the consistency makes the code easier to read.

src/LogFileReader.cpp
```diff
@@ -67,19 +71,38 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    }
+
+    batteryData.voltage = lines[1].toDouble(&ok);
+    batteryData.current = lines[2].toDouble(&ok);
```
enochtsang: It would be better to check the `ok` variable after each of these lines, less repeated code.

src/LogFileReader.cpp
```diff
@@ -67,19 +71,38 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    {
+        return false;
+    }
 
-    // This is here to the compiler happy. Otherwise the compile
-    // will have an error warning about an unused variable. Remove this
-    // when you use it.
-    Q_UNUSED(line);
-    Q_UNUSED(batteryData);
+    for (int i = 1; i<=2; i++)
```
enochtsang: It would be much more standard to use `for (int i = 0; i < 2; i++)`. It is not common to use `<=` for the conditional statements in for loops.

src/LogFileReader.cpp
```diff
@@ -67,19 +71,38 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    }
+    else
+    {
+        return false;
+    }
+
+    if (lines.count() !=    3)
+    {
+        return false;
+    }
```
enochtsang: Doing the error checking together will make for shorter and easier to read code. So you can just use
```c++
if (!Time.isValid() || lines.count() != 3)
{
    return false;
}
```

src/LogFileReader.cpp
```diff
@@ -67,19 +71,38 @@ bool LogFileReader::readAll(const QString& fileName)
...
-
     return true;
 }
 
 // File input is a csv file in the format of hh:mm:ss:zzz, voltage, current
 bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) const
 {
-    // TODO implement this first
+    QStringList lines = line.split(COMMA);
+    QTime Time = QTime::fromString(lines[0], TIMEFORMAT);
```
enochtsang: `Time` should be lowercase

src/BatteryStateOfChargeService.cpp
```diff
@@ -1,40 +1,46 @@
...
  *
  *  For further contact, email <software@calgarysolarcar.ca>
  */
-
+#include <QDebug>
+#include "BatteryData.h"
 #include "BatteryStateOfChargeService.h"
-
+#include <QTime>
+#include <math.h>
```
enochtsang: I don't think this is being used.

src/BatteryStateOfChargeService.cpp
```diff
@@ -1,40 +1,46 @@
...
 namespace
 {
     const double BATTERY_AMP_HOUR_CAPACITY = 123.0;
+    const int SECONDS_TO_HOURS = 3600;
+    const int SECONDS_TO_MINUTES = 60;
+    const int SECONDS_TO_MILISECONDS = 1000;
```
enochtsang: I think a few of these are not being used.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +47,49 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 double BatteryStateOfChargeService::totalAmpHoursUsed() const
 {
-    return 0.0;
+    return totalAmpUsed_;
 }
 
 bool BatteryStateOfChargeService::isCharging() const
 {
-    return false;
+    if (newCurrent_ <= 0)
```
enochtsang: If the `newCurrent_` is equal to `0` I would say that it is probably not charging haha.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +47,49 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+        return false;
+    }
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
-}
+    double newTimeSecs = QTime(0,0,0).secsTo(newTime_);
+    double TimeElapsed =(BATTERY_AMP_HOUR_CAPACITY / fabs(newCurrent_)) * SECONDS_TO_HOURS;
```
enochtsang: Local variables should be lower case

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +47,49 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+}
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+    QTime oldTime = QTime();
+    if (oldTime.isNull() == true)
+    {
+        oldTime = QTime(0,0,0);
+
```
enochtsang: This section of code seems odd, you set `oldTime` as `QTime()` then check it in a conditional? You must know what the status of `oldTime` is since you set it right before this.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +47,49 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+        oldTime = newTime_;
+    }
+
+    double oldCurrent = newCurrent_;
+    newTime_ = batteryData.time;
+    newCurrent_ = batteryData.current;
+
+    double TimeUsed = newTime_.msecsTo(oldTime);
+    double averageCurrent = (newCurrent_ + oldCurrent) / 2;
+    totalAmpUsed_ = totalAmpUsed_ + averageCurrent  / (TimeUsed / SECONDS_TO_MILISECONDS);
```
enochtsang: You can use the `+=` operator here. So example:
```c++
x = x +3;
```

Is the equivalent of
```c++
x += 3;
```

src/BatteryStateOfChargeService.h
```diff
@@ -42,8 +42,11 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
-    // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
-
-    // Recommend adding necessary variables here
+    double newCurrent_;
+    double totalCurrent_;
+    QTime estimatedTime_;
+    QTime newTime_;
+    double totalAmpUsed_;
+    double initialAmpUsed_;
```
enochtsang: Does this really need to be a private member?

src/BatteryStateOfChargeService.h
```diff
@@ -42,8 +42,11 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     void addData(const BatteryData& batteryData);
 
 private:
-    // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
-
-    // Recommend adding necessary variables here
+    double newCurrent_;
+    double totalCurrent_;
```
enochtsang: Is this still being used?

src/LogFileReader.cpp
```diff
@@ -67,19 +71,35 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    }
 
-    // This is here to the compiler happy. Otherwise the compile
-    // will have an error warning about an unused variable. Remove this
-    // when you use it.
-    Q_UNUSED(line);
-    Q_UNUSED(batteryData);
-    return true;
+
+    if((lines[1].toDouble(&ok) == ok) || (lines[2].toDouble(&ok) == ok))
```
enochtsang: Might want to check the logic here, you're comparing doubles to booleans here, I don't think that's what you're intending to do.

enochtsang:
Make sure you watch your white spacing as well, you got some extra lines in there that need to be deleted.

src/LogFileReader.cpp
ikazuka36: lines[1].toDouble(&ok) would convert that variable into a double, and if the operation fail (not a number) it would convert that variable into 0 with a boolean value of ok. That's why in that conditional sentence, i'm setting if there are booleans values which mean the convert failed and should be return false

src/LogFileReader.cpp
```diff
@@ -67,19 +72,27 @@ bool LogFileReader::readAll(const QString& fileName)
...
-    // This is here to the compiler happy. Otherwise the compile
-    // will have an error warning about an unused variable. Remove this
-    // when you use it.
-    Q_UNUSED(line);
-    Q_UNUSED(batteryData);
-    return true;
+    QStringList lines = line.split(COMMA);
+    QTime time = QTime::fromString(lines[0], TIMEFORMAT);
+    bool notok = false;
+    if (time.isValid() && lines.count() == 3 && (lines[1].toDouble(&notok) != notok) && (lines[2].toDouble(&notok) != notok))
```
enochtsang: Hmm, not quite what I was intending. As a style thing you want to avoid very long lines, doing 
```c++
if (time.isValid() && lines.count() == 3 && 
    lines[1].toDouble(&notok) != notok && 
    lines[2].toDouble(&notok) != notok)
```
Would look better.

But more importantly, I want you to think about
`lines[1].toDouble(&notok) != notok` and why it won't work.

src/LogFileReader.cpp
```diff
@@ -67,19 +72,27 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    }
+    else
+    {
+        return false;
+    }
 }
+
+
+
+
```
enochtsang: Extra white space here.

src/BatteryStateOfChargeService.cpp
```diff
@@ -1,40 +1,44 @@
...
-
+#include <QDebug>
+#include "BatteryData.h"
 #include "BatteryStateOfChargeService.h"
-
 namespace
 {
     const double BATTERY_AMP_HOUR_CAPACITY = 123.0;
+    const int SECONDS_TO_HOURS = 3600;
+    const int SECONDS_TO_MILISECONDS = 1000;
```
enochtsang: Typo on `MILISECONDS`, should be milliseconds.

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +47,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    {
+        return false;
+    }
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    double seconds = QTime(0,0,0).secsTo(newTime_);
+    double timeelapsed =(BATTERY_AMP_HOUR_CAPACITY / fabs(newCurrent_)) * SECONDS_TO_HOURS;
```
enochtsang: Sorry I guess I should've been clearer about naming. Local variables should follow camelCase convention. Where the first letter of each word is capitalized, except for the first word. Some other types of naming conventions are cap case and snake case. Here are some examples.

`CAP_CASE` mainly for constants.
`IRS_TIME_DONE`
`CONST`

`snake_case` is a standard for python and is becoming more and more prevalent in C++
`an_example`
`irs_time`

src/BatteryStateOfChargeService.cpp
```diff
@@ -43,21 +47,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    QTime estimatedtime = QTime(0,0,0).addSecs(seconds);
+    return estimatedtime;
 }
-
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
-    // Update your variables here.
+    QTime oldtime = QTime();
+    oldtime = newTime_;
```
enochtsang: These 2 lines here can be compressed into one.

src/LogFileReader.cpp
ikazuka36: I've made a comment on that but you probably missed it. 

lines[1].toDouble(&notok) would convert that line[1] into a double, and if the operation fail (not a number) it would convert line[1] into 0 with a boolean value of notok. That's why in that conditional sentence, i'm setting if there are not booleans values which mean the convert was successful then update the data

enochtsang: This is not quite as clear. I'd recommend converting first, then checking `notok`. So you only do one conversion.

enochtsang:
Looks pretty good now, hope you've learnt a lot from this assignment. I'll test it tomorrow then we can close this PR and we get you started on a project.

enochtsang:
I get the error

```
BatteryStateOfChargeService.cpp: In member function ‘virtual QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const’:
BatteryStateOfChargeService.cpp:68:70: error: ‘fabs’ was not declared in this scope
     double timeElapsed =(BATTERY_AMP_HOUR_CAPACITY / fabs(newCurrent_)) * SECONDS_TO_HOURS;
                                                                      ^
make: *** [../release/.obj/BatteryStateOfChargeService.o] Error 1
```

When compiling.

enochtsang:
Finished

PR #13, opened 2016-08-16
Fill parser

bill-luu:
The purpose of this PR is to fill in the logfilereader class, as well has make some changes to the README and add comments that clarify the purpose of the exercise.


src/BatteryStateOfChargeService.h
```diff
@@ -36,14 +36,20 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     virtual ~BatteryStateOfChargeService();
 
     double totalAmpHoursUsed() const;
+    /*Promises: Returns the total amp hours used*/
+
     bool isCharging() const;
+    /*Promises: Returns through if the battery is being charged, false if not.*/
+
     QTime timeWhenChargedOrDepleted() const;
+    /*Promises: Return the time remaining to when the battery is charged or depleted*/
```
enochtsang: Let's not use the "promises", just an explanation. Also generally comments go above a function.

src/LogFileReader.cpp
```diff
@@ -32,6 +32,13 @@
 
 #include "LogFileReader.h"
 
+namespace
+{
+    const QString STRING_TIME_FORMAT= "hh:mm:ss.zzz";
+    const QString BatData_Delimiter= ", ";
```
enochtsang: all caps

src/LogFileReader.cpp
```diff
@@ -74,12 +81,28 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    batteryData.current = sections.at(2).toDouble(&currentOK);
+
+    if(voltageOK && currentOK && batteryData.time.isValid())
+    {
+       return true;
+    }
+    else
+    {
+       return false;
+    }
```
enochtsang: IIRC, we said error checking would still be on them. Only file parsing without error checking is provided.

src/LogFileReader.cpp
```diff
@@ -71,15 +78,20 @@ bool LogFileReader::readAll(const QString& fileName)
...
 }
 
-// File input is a csv file in the format of hh:mm:ss:zzz, voltage, current
+/* File input is a csv file in the format of hh:mm:ss:zzz, voltage, current
+ * Need to implement error checking for the correct number of values and
+ * that the conversion from string to double is sucessful.*/
 bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) const
 {
-    // TODO implement this first
+    QStringList sections = line.split(BatData_Delimiter);
```
enochtsang: Doesn't compile, you need to update this variable.

src/LogFileReader.cpp
```diff
@@ -71,15 +78,20 @@ bool LogFileReader::readAll(const QString& fileName)
...
+
+    QString timeString = sections.at(0);
+    batteryData.time = QTime::fromString(timeString,STRING_TIME_FORMAT);
```
enochtsang: space after comma

PR #25, opened 2016-11-16
Modified some comments to clarify/give hints on some obscure concepts

bill-luu:
@enochtsang @cheunggg Please take a look at this, and see if there are any other components/concepts that may be in need of explaining! I'll hand Gordon the merge rights for this.


enochtsang:
LGTM

PR #27, opened 2016-11-26
Sarah

sarahlam1009:
Finished the first basic assignment.

### DIFF
--- a/Basic-Recruit-Training/Person.cpp
+++ b/Basic-Recruit-Training/Person.cpp
@@ -1,0 +1,43 @@
+#include "Person.h"
+
+Person :: Person ()
+{
+
+}
+
+Person :: Person(const std::string& name, int age)
+{
+    name_ = name;
+    age_ = new int(age);
+}
+
+const std::string& Person :: getName () const
+{
+    return name_;
+}
+
+int Person :: getAge () const
+{
+    return *age_;
+}
+
+void Person :: setAge (int newAge)
+{
+   delete(age_);
+   age_ = new int(newAge);
+}
+
+static int Person :: combinedAge(Person** x, int size)
+{
+
+
+
+}
+
+
+Person :: ~Person ()
+{
+
+}
+
+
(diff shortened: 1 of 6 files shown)


Basic-Recruit-Training/Person.cpp
```diff
@@ -0,0 +1,62 @@
+#include "Person.h"
+
+Person :: Person ()
+{
+
```
cheunggg: Even though we set the member variables of Person in the main code, the default constructor should still assign the member variables to some default value, usually these are null values

Basic-Recruit-Training/Person.cpp
```diff
@@ -0,0 +1,62 @@
...
+   for (int counter = 0; counter < size; counter++ )
+   {
+      ageCombined = ageCombined + (x[counter])->getAge();
+   }
+   return ageCombined;
+}
+
+void Person :: birthday(Person& x)
+{
+    x.setAge (x.getAge() + 1);
```
cheunggg: No spaces should separate the name of the function and the brackets, this makes it look like there are two things, when in fact it is actually one

Basic-Recruit-Training/Person.cpp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+Person :: ~Person ()
+{
+    delete(age_);
+}
+
+
```
cheunggg: Have only one line at the end of files

Basic-Recruit-Training/Person.cpp
```diff
@@ -0,0 +1,62 @@
...
+void Person :: printInfo ()
+{
+    std :: cout << "The Person's Name is:" << getName() << std :: endl;
+    std :: cout << "The Person's Age is :" << getAge() << std :: endl;
+}
+
+void Person :: setAge (int newAge)
+{
+   delete(age_);
```
cheunggg: There is a problem with using delete here

Basic-Recruit-Training/Person.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+
+int Person :: getAge () const
+{
+    return *age_;
+}
+
+void Person :: printInfo ()
+{
+    std :: cout << "The Person's Name is:" << getName() << std :: endl;
```
cheunggg: Good

PR #33, opened 2017-01-05
Michael verwaayen advanced training

michaelverwaayen:


Advanced-Recruit-Training/src/BatteryData.h
```diff
@@ -4,6 +4,10 @@
 
 struct BatteryData
 {
+
+
+
+
```
enochtsang: Unnecessary.

Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -4,7 +4,7 @@
 #include "BatteryStateDisplayService.h"
 #include "I_BatteryDataSource.h"
 #include "I_BatteryStateOfChargeService.h"
-
+#include "BatteryStateOfChargeService.h"
```
enochtsang: Shouldn't be necessary, you would only need this header if you were instantiating it here, but you only need the interface type `I_BatteryStateOfChargeService`

Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -4,7 +4,7 @@
 #include "BatteryStateDisplayService.h"
 #include "I_BatteryDataSource.h"
 #include "I_BatteryStateOfChargeService.h"
-
```
enochtsang: Includes should be buffered by a new line.

Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -24,10 +24,12 @@ BatteryStateDisplayService::~BatteryStateDisplayService()
...
+/*
     QTextStream(stdout) << "Voltage: " << batteryData.voltage
         << " Current: " << batteryData.current
-        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed() << endl;
-
+        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed()
+    //QString text =QTime::currentTime().toString();
+        << "TIME UNTIL BATTERY FULLY CHARGED" << batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString()<<endl;
+*/
```
enochtsang: In general, commented code should be deleted in a PR. That's what version control is for.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,4 +1,6 @@
 #include "BatteryStateOfChargeService.h"
+#include "BatteryData.h"
+#include <QTextStream>
```
enochtsang: Is this necessary?

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +17,109 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    if(avg <= 0)
+    {
+        avg= -avg;
+    }
+    double averageMin=0; //current/60 will be current per minutes
+    double averageSec=0; //average min /60 will be current per seconds
+    int hours=0;   //variable to multiply by 3600
+    int minutes=0; //variable to multiply by 60
+    int seconds=0; //variable to multiply by 1
+    if(isCharging())
```
bill-luu: This is a unique solution, however using these two loops for all these values is pretty intense :p. There's a two line, mathematical solution that I want you to use instead to determine how many milliseconds are left until the battery is charged or depleted.

Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -27,7 +26,6 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
 
     QTextStream(stdout) << "Voltage: " << batteryData.voltage
         << " Current: " << batteryData.current
-        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed() << endl;
-
-    // TODO: Print out time until the battery is fully charged or depleted.
+        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed()
+        << "TIME UNTIL BATTERY FULLY CHARGED" << batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString()<<endl;
```
bill-luu: Here is a sample output:
`Voltage: 3.11293 Current: 0.530217 Total Ah used: 5.47108e+249TIME UNTIL BATTERY FULLY CHARGED02:12:02`
There are 2 errors here, can you figure out what they are?

bill-luu:
You should take a look at the style guide to learn more about what we're looking for in coding style.
This can be found in `Recruit-Resources/Development-Guide/Style-Guide/`. We use a linter called AStyle to do auto-format our code, you should take a look at the `astylerc` file in `/Style-Guide` to see what we look for in style.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
```diff
@@ -15,14 +15,19 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
 private:
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
-
+    double current;
+    double voltage;
+    QTime previousTime;
+    QTime currentTime;
+    double currentAmpHours;
+    double timeDifferenceStamp;
```
enochtsang: Private member should have underscores

Advanced-Recruit-Training/src/I_BatteryStateOfChargeService.h
```diff
@@ -13,4 +13,6 @@ class I_BatteryStateOfChargeService
     virtual QTime timeWhenChargedOrDepleted() const = 0;
 
     virtual void addData(const BatteryData& batteryData) = 0;
+//private:
+//    int counter=0;
```
enochtsang: No need for commented code. I would just like to make an additional note as well, this is an interface class as denoted by the `I_`, there should never be any private members in an interface.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -34,38 +32,43 @@ bool LogFileReader::readAll(const QString& fileName)
     {
         QString line = input.readLine();
         BatteryData batteryData;
-        if (!parseLine(line, batteryData))
+        if (!parseLine(line, batteryData))  //! taken before parseline out
```
enochtsang: This function wasn't supposed to be edited at all. This goes for in the future too, try not to edit code that isn't related to the feature being currently worked on, doing so messes with diffs and stuff in version control.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -34,38 +32,43 @@ bool LogFileReader::readAll(const QString& fileName)
...
 bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) const
 {
+    bool ok;
     QStringList sections = line.split(BATDATA_DELIMITER);
-
+    if(sections.size() != 3)
+    {
+        return false;
+    }  
+    if(sections.contains("[Aa-Zz]")) //make sure it doesnt contain alphabetical characters
```
enochtsang: I don't think the function can handle expressions like this. If it does, can I get a link to the docs that describes this? I see this in the docs 
http://doc.qt.io/qt-5/qstringlist.html#contains

Advanced-Recruit-Training/src/main.cpp
```diff
@@ -1,7 +1,6 @@
 #include <QCommandLineParser>
 #include <QCoreApplication>
 #include <QTimer>
-
```
enochtsang: This file shouldn't be changed.

Basic-Recruit-Training/main.cpp
```diff
@@ -19,10 +19,16 @@ void compareAnswers(const std::string myAnswer, const std::string correctAnswer)
 // Swaps a & b.
 void mySwap(int& a, int& b)
 {
+    int temp=b;
+    a=b;
+    b=temp;
+    return a,b
```
enochtsang: This is a void function, returning these arguments doesn't do anything. C++ also does not support returning multiple values like this (I think that was the intent), so actually I don't think this compiles?

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +17,109 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
 
 double BatteryStateOfChargeService::totalAmpHoursUsed() const
```
enochtsang: Well written Bill :+1:

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +17,66 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    {
+        return false;
+    }
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    double initialValueOfCharge = initialStateOfChargePercent_ * BATTERY_AMP_HOUR_CAPACITY; //initial state in miliseconds
+    double averageIncreaseValue = totalAmpHoursUsed() * 3.6e+6 ; //the increase in charge
```
enochtsang: This 3.6e+6 is what we call a magic number. Abstract it out to a named variable.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +17,66 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    else
+    {
+        return false;
+    }
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    double initialValueOfCharge = initialStateOfChargePercent_ * BATTERY_AMP_HOUR_CAPACITY; //initial state in miliseconds
```
enochtsang: I think a better would be `initialAmpHours`

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +17,66 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+        if(currentTime.toString() == NULL)
+        {
+            previousTime = QTime(0,0,0);
+        }
+        else
+        {
+            previousTime = currentTime;
+        }
+        currentTime = batteryData.time;
+        timeDifferenceStamp = currentTime.toString( "hhmmss.zzz").toDouble() - previousTime.toString( "hhmmss.zzz" ).toDouble();
```
enochtsang: You can't subtract the time stamps.

1:22:00 - 0:23:00 makes for an actual time change of 0:59:00, but you would get 0:99:00. You should convert them to seconds.

Consider what unit `timeDifferenceStamp` is.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +17,66 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
-    Q_UNUSED(batteryData);
-    // This is where you can update your variables
-    // Hint: There are many different ways that the totalAmpHoursUsed can be updated
-    // i.e: Taking a running average of your data values, using most recent data points, etc.
+    {
+        voltage = batteryData.voltage;
+        current = batteryData.current; //get the current value
+        if(currentTime.toString() == NULL)
+        {
+            previousTime = QTime(0,0,0);
```
enochtsang: You can't just set a time of 0, because the next time could be way in the future but that amount of time hasn't actually elapsed. You want to consider not changing the ampHours data when you only have one data point.

Setting a time of 0 for the first data point doesn't work, because that could create a huge time difference between the first data point and the previous time. This will greatly skew the calculations. One possibility is not changing the ampHours data when only one data point is available.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
enochtsang: Also, voltage, timeDifferenceStamp, and currentTime don't need to be private members. Their longevity is only one call in `addData`. The scope of all variables should be as small as possible.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +17,66 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+        }
+        else
+        {
+            previousTime = currentTime;
+        }
+        currentTime = batteryData.time;
+        timeDifferenceStamp = currentTime.toString( "hhmmss.zzz").toDouble() - previousTime.toString( "hhmmss.zzz" ).toDouble();
+
+        currentAmpHours = currentAmpHours + current/timeDifferenceStamp;
```
enochtsang: An amphour calculation can't be done with only data point. So finding the change in time is correct, but a change in current must be taken into account as well. 

Take for example these two data points:
t=0s, I=100mA
t=10s, I=0mA

By the approach given here, the calculation be we change in time (10s), and the last current (0mA). This equates to a change in 0 amp hours, but it is probably not reasonable to assume that in that whole 10s that no current was being used.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +17,66 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+        //This was moved form above
+        isCharging(); //call to change it
+        timeWhenChargedOrDepleted();
```
enochtsang: These two functions are const, meaning they shouldn't change the state of the class at all. This should all be done in addData. The mentality behind this is that `isCharging()` and `timeWhenChargedOrDepleted()` are getter functions. They could be called 100 times or 0 times in between every addData function, but that knowledge is abstracted out of the this class.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
```diff
@@ -15,14 +15,19 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     bool isCharging() const;
 
     /*Return the time remaining to when the battery is charged or depleted*/
-    QTime timeWhenChargedOrDepleted() const;
+    QTime timeWhenChargedOrDepleted()const;
```
enochtsang: This looks like a syntax error.

enochtsang:
Closing until further work is done.

PR #36, opened 2017-02-11
Progress on Advanced Recruit Training

louisj381:


Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,36 +1,70 @@
 #include "BatteryStateOfChargeService.h"
+#include <QDebug>
```
bill-luu: Doesn't like you're using `qDebug` in this file, so this should be dropped.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,36 +1,70 @@
...
 
+
 BatteryStateOfChargeService::~BatteryStateOfChargeService()
 {
 }
 
 double BatteryStateOfChargeService::totalAmpHoursUsed() const
 {
-    return 0.0;
+    return initialAhUsed_ + chargeAh_;
```
bill-luu: It makes more sense to have a single variable `totalAmpHoursUsed_` that's calculated in `addData` rather than doing an addition here.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,36 +1,70 @@
...
 }
 
 bool BatteryStateOfChargeService::isCharging() const
 {
-    return false;
+    if (current_ < 0)
+       return true;
+        return false;
```
bill-luu: Even though it wouldn't do much, adding this to an `else` statement is better practice.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
```diff
@@ -23,6 +23,15 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
 private:
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
+    double initialAhUsed_;
+    double chargeAh_;
+    double averageCurrent_;
+    double totalCurrent_ = 0;
```
bill-luu: Typically values aren't assigned in the header file, if you need to assign a starting value, use the constructor!

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
```diff
@@ -23,6 +23,15 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
+    int counter_ = 1;
```
bill-luu: As above.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
```diff
@@ -23,6 +23,15 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
 private:
     // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
+    double initialAhUsed_;
+    double chargeAh_;
+    double averageCurrent_;
```
bill-luu: Consider the lifetime of this variable: Its value isn't used outside of `addData` and the value doesn't carry over into other calls of `addData`. With that in mind, it's probably best for it to be a local variable.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,36 +1,70 @@
...
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    QTime time(0,0,0,0);
+    int ms_remaining;
+      if (isCharging() == true)  {
+        ms_remaining = totalAmpHoursUsed()/averageCurrent_ *3.6E6;
+    }   else    {
+        ms_remaining = (BATTERY_AMP_HOUR_CAPACITY - totalAmpHoursUsed())/averageCurrent_ *3.6E6;
```
bill-luu: See my other comments on magic numbers

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,36 +1,70 @@
...
+        ms_remaining *= -1;
```
bill-luu: Instead of multiplying by -1, take the absolute value instead.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,36 +1,70 @@
...
+        return false;
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    QTime time(0,0,0,0);
+    int ms_remaining;
+      if (isCharging() == true)  {
+        ms_remaining = totalAmpHoursUsed()/averageCurrent_ *3.6E6;
```
bill-luu: Leave a space between arithmetic operators, so `totalAmpHoursUsed() / averageCurrent_ * 3.6E6`

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,36 +1,70 @@
...
 }
 
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
-    // This is where you can update your variables
+
+    QTime updated_time;
+
+    double DeltaMseconds = batteryData.time.msec() - Previous_milliseconds_.msec();
```
bill-luu: One more point to consider, `DeltaMseconds` relies on the value of `Previous_milliseconds.msec()`, however when on the very first run, `Previous.millseconds.msec()` doesn't have a value at all, so the results can be skewed. I'll let you try and figure out how to work around this.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -52,36 +52,39 @@ bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) con
...
+    {
         qDebug() << "sections length is greater than 3";
         return false;
     }
 
     QString timeString = sections.at(0);
     batteryData.time = QTime::fromString(timeString, STRING_TIME_FORMAT);
 
-    bool double_conversion;
+    bool doubleConversionV;
```
bill-luu: No need to abbreviate `Voltage`, or `Current`

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -52,36 +52,39 @@ bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) con
...
 
-    if (double_conversion == false) {
-        qDebug() << "double_conversion unsuccessful";
-        return false;
-}
-    batteryData.current = sections.at(2).toDouble(&double_conversion);
+    batteryData.current = sections.at(2).toDouble(&doubleConversionC);
 
-    if (double_conversion == false) {
+    if (doubleConversionV == false || doubleConversionC == false)
```
bill-luu: Going with what I said earlier, can you guess what the equivalent statement would be?

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
```diff
@@ -23,6 +24,13 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
...
-    // Add your necessary variables here
+    double totalAmpHoursUsed_;
+    double averageCurrent_; //
+    double totalCurrent_;
+    int counter_;
+    double hours_remaining_;
+    double current_;
+    int initialAhUsed_;
+    QTime initialTime_;
+    QTime previousMilliseconds_;
```
bill-luu: `previousMilliseconds` isn't really a great name, it implies that it returns the value of milliseconds of the previous time, while actually it just contains the previous time.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +26,63 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    int deltaMseconds;
+    double chargeAh;
+    QTime msTime(0,0,0,0);
+
+    if (previousMilliseconds_.isNull())
+    {
+        previousMilliseconds_ = initialTime_.addMSecs(batteryData.time.msec());
+    }
+
+    msTime = msTime.addMSecs(batteryData.time.msec());
```
bill-luu: This is basically taking `batteryData.time`, converting it to milliseconds, and then assigning `msTime` to a QTime that is `batteryData.time.msec()` milliseconds ahead. Since the initial value of `msTime` is 0:00:00.000, let's just assign `msTime ` to `batteryData.time`!

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +26,63 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+    deltaMseconds = abs(msTime.msecsTo(previousMilliseconds_));
+
+    double deltaHours = deltaMseconds * HOUR_CONVERSION;
+
+    previousMilliseconds_ = initialTime_.addMSecs(batteryData.time.msec());
```
bill-luu: Similarly to above, we can just set `previousMilliseconds_ = batteryData.time` and we don't need the `initialTime` and `placeholder` variables.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +26,63 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
 }
 
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
-    // This is where you can update your variables
+    int deltaMseconds;
+    double chargeAh;
+    QTime msTime(0,0,0,0);
```
bill-luu: This variable name is also a bit misleading, since it is `batteryData.time`, not just the milliseconds. Perhaps to differentiate it from the previous time, name it something along the lines of `presentTime`

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +26,63 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+    previousMilliseconds_ = initialTime_.addMSecs(batteryData.time.msec());
+
+    current_ = batteryData.current;
+
+    totalCurrent_ = batteryData.current + totalCurrent_;
+
+    averageCurrent_ = totalCurrent_ / counter_;
+
+    counter_++;
```
bill-luu: Put this increment at the beginning, and set the initial value to zero. While functionally the same, it makes more sense to increment the counter of data only when new data is coming in, not before.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
louisj381: I am not sure how you want me to adjust this, if I assign msTime (now presentTime) to batteryData.time, the calculations are dis-proportionally of, as the number of miliseconds that the first instance of batteryData.time has is very large, this was the issue I was having earlier.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
louisj381: I thought it does contain the milliseconds of the previous time, the way I have it set up.

Advanced-Recruit-Training/src/main.cpp
```diff
@@ -8,7 +8,7 @@
 
 int main(int argc, char *argv[])
 {
-    QCoreApplication app(argc, argv);
+    QCoreApplication app(argc, argv);   //shell for main function
```
bill-luu: Comments here still need to be removed.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
bill-luu: Ah wait that's my mistake, I was wrong on the `time.msec`, you're right it does just take the millisecond part. However see my overall comment.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -64,18 +64,20 @@ void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
     double chargeAh;
     QTime presentTime(0,0,0,0);
 
+    int Milliseconds = QTime(0, 0, 0, 0).msecsTo(batteryData.time);
```
bill-luu: Variable names should start with a lowercase letter.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -37,34 +36,55 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    bool doubleConversionVoltage;
+    bool doubleConversionCurrent;
 
-    batteryData.current = sections.at(2).toDouble();
+    batteryData.voltage = sections.at(1).toDouble(&doubleConversionVoltage);
 
-    return true;
+    batteryData.current = sections.at(2).toDouble(&doubleConversionCurrent);
+
+    if (doubleConversionVoltage == false || doubleConversionCurrent == false)
```
bill-luu: I think I mentioned this before, but try formatting this like you did for `if(batteryData.time.isValid())` below.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +26,66 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    QTime presentTime(0,0,0,0);
+
+    int Milliseconds = QTime(0, 0, 0, 0).msecsTo(batteryData.time);
+
+    if (previousMilliseconds_.isNull())
+    {
+        previousMilliseconds_ = initialTime_.addMSecs(Milliseconds);
+    }
+
+    presentTime = presentTime.addMSecs(Milliseconds);
```
bill-luu: This line and the initialization of `presentTime` can be replaced by assigning `presentTime` to `batteryData.time`

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +26,66 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    counter_++;
+    int deltaMseconds;
+    double chargeAh;
+    QTime presentTime(0,0,0,0);
+
+    int Milliseconds = QTime(0, 0, 0, 0).msecsTo(batteryData.time);
+
+    if (previousMilliseconds_.isNull())
+    {
+        previousMilliseconds_ = initialTime_.addMSecs(Milliseconds);
```
bill-luu: Instead of adding the milliseconds between 0 and `batteryData.time` to `initialTime_`, and then assigning it to `previousMilliseconds`, we can now just assign `previousMilliseconds` to `batteryData.time`, which is equivalent, but more cleaner.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +26,66 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    }
+
+    presentTime = presentTime.addMSecs(Milliseconds);
+
+    deltaMseconds = abs(presentTime.msecsTo(previousMilliseconds_));
+
+    double deltaHours = deltaMseconds * MS_TO_HOURS_CONVERSION;
+
+    previousMilliseconds_ = initialTime_.addMSecs(Milliseconds);
```
bill-luu: As above.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +26,66 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
-    // This is where you can update your variables
+    counter_++;
+    int deltaMseconds;
+    double chargeAh;
+    QTime presentTime(0,0,0,0);
+
+    int Milliseconds = QTime(0, 0, 0, 0).msecsTo(batteryData.time);
```
bill-luu: Variables should start with a lower case letter.

Advanced-Recruit-Training/src/main.cpp
bill-luu: Still need to be removed!

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -37,34 +36,55 @@ bool LogFileReader::readAll(const QString& fileName)
...
- * Need to implement error checking for the correct number of values and
- * that the conversion from string to double is sucessful.*/
+
 bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) const
 {
     QStringList sections = line.split(BATDATA_DELIMITER);
 
+    if (sections.length() < COLUMNS)
+    {
+        qDebug() << "sections length is less than 3";
```
bill-luu: We don't really need these Debug messages, since readAll() already prints out one when `parseLine` returns false.
Also because readAll()'s message will print out after, so reading `batDataFailure.csv` would look like this:
`Voltage: 3.6 Current: -12.1075 Total Ah used: 61 Time remaining until the Battery is Charged: 18:57:42.494`
`time is invalid`
`Error while parsing "07:00,04.926, 3.596829, -2.408174"`
Which looks a bit funky, since it looks like it's saying that the first line is invalid, when it's really the next one.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -37,34 +36,55 @@ bool LogFileReader::readAll(const QString& fileName)
...
 
-/* File input is a csv file in the format of hh:mm:ss:zzz, voltage, current
- * Need to implement error checking for the correct number of values and
- * that the conversion from string to double is sucessful.*/
+
 bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) const
 {
     QStringList sections = line.split(BATDATA_DELIMITER);
 
+    if (sections.length() < COLUMNS)
```
bill-luu: As a corollary of my comment below, we can just check if sections.length() is equal to `COLUMNS`.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -37,34 +36,55 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    {
+        qDebug() << "sections length is less than 3";
+        return false;
+    }
+
+    else if (sections.length() > COLUMNS)
+    {
+        qDebug() << "sections length is greater than 3";
```
bill-luu: See above.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -37,34 +36,55 @@ bool LogFileReader::readAll(const QString& fileName)
...
 
-    batteryData.current = sections.at(2).toDouble();
+    batteryData.voltage = sections.at(1).toDouble(&doubleConversionVoltage);
 
-    return true;
+    batteryData.current = sections.at(2).toDouble(&doubleConversionCurrent);
+
+    if (!doubleConversionVoltage || !doubleConversionCurrent)
+    {
+        qDebug() << "double_conversion unsuccessful";
```
bill-luu: See above.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +26,63 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+      {
+        ms_remaining = totalAmpHoursUsed() / averageCurrent_ * HOURS_TO_MS_CONVERSION;
+      }
+
+      else
+      {
+        ms_remaining = (BATTERY_AMP_HOUR_CAPACITY - totalAmpHoursUsed()) / averageCurrent_ * HOURS_TO_MS_CONVERSION;
+      }
+
+    QTime timeWhenChargedOrDepleted = time.addMSecs(ms_remaining);
```
enochtsang: I would've used: `return QTime(0, 0, 0, 0).addMSecs(ms_remaining)`

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +26,63 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+    current_ = batteryData.current;
+
+    totalCurrent_ = batteryData.current + totalCurrent_;
+
+    averageCurrent_ = totalCurrent_ / counter_;
+
+    chargeAh =  averageCurrent_ * deltaHours;
+
+    totalAmpHoursUsed_ = initialAhUsed_ + chargeAh;
```
enochtsang: Don't need this many extra lines.

Group blocks of logic together.

Also `totalAmpHoursUsed_ = initialAhUsed_ + chargeAh;` has a bug

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +26,63 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    return timeWhenChargedOrDepleted;
+
 }
 
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
-    // This is where you can update your variables
+    counter_++;
+    int deltaMseconds;
```
enochtsang: Move this to where you use it.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +26,63 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+    previousMilliseconds_ = batteryData.time;
+
+    current_ = batteryData.current;
+
+    totalCurrent_ = batteryData.current + totalCurrent_;
+
+    averageCurrent_ = totalCurrent_ / counter_;
+
+    chargeAh =  averageCurrent_ * deltaHours;
```
enochtsang: Much better named `deltaAh_`

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +26,63 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+    deltaMseconds = abs(presentTime.msecsTo(previousMilliseconds_));
+
+    double deltaHours = deltaMseconds * MS_TO_HOURS_CONVERSION;
+
+    previousMilliseconds_ = batteryData.time;
+
+    current_ = batteryData.current;
+
+    totalCurrent_ = batteryData.current + totalCurrent_;
```
enochtsang: Use a running average compared to a total average like you have here.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
```diff
@@ -1,28 +1,35 @@
...
 private:
-    // Convention is to have an _ after private members
     const double initialStateOfChargePercent_;
-
-    // Add your necessary variables here
+    double totalAmpHoursUsed_;
+    double averageCurrent_; //
+    double totalCurrent_;
+    int counter_;
+    double hours_remaining_;
```
enochtsang: Fix this naming

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
```diff
@@ -1,28 +1,35 @@
...
+    double current_;
+    int initialAhUsed_;
```
enochtsang: Doesn't need to be stored.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
```diff
@@ -1,28 +1,35 @@
...
+    QTime initialTime_;
```
enochtsang: Unused

Advanced-Recruit-Training/src/I_BatteryStateOfChargeService.h
```diff
@@ -13,4 +14,5 @@ class I_BatteryStateOfChargeService
     virtual QTime timeWhenChargedOrDepleted() const = 0;
 
     virtual void addData(const BatteryData& batteryData) = 0;
+
```
enochtsang: This file didn't need to be edited.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -37,34 +36,46 @@ bool LogFileReader::readAll(const QString& fileName)
         if (!parseLine(line, batteryData))
         {
             qDebug() << "Error while parsing" << line;
-            // return false;
+             return false;
```
enochtsang: This must've caused alot of headaches for you.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -37,34 +36,46 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    {
+        return false;
+    }
+
     QString timeString = sections.at(0);
     batteryData.time = QTime::fromString(timeString, STRING_TIME_FORMAT);
 
-    batteryData.voltage = sections.at(1).toDouble();
+    bool doubleConversionVoltage;
+    bool doubleConversionCurrent;
```
enochtsang: `voltageOk`
`currentOk`

Would be better

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -37,34 +36,46 @@ bool LogFileReader::readAll(const QString& fileName)
...
 
-    return true;
+    batteryData.current = sections.at(2).toDouble(&doubleConversionCurrent);
+
+    if (!doubleConversionVoltage || !doubleConversionCurrent)
+    {
+        return false;
+    }
 
+     if (batteryData.time.isValid())
```
enochtsang: Use braces

Advanced-Recruit-Training/src/main.cpp
```diff
@@ -43,7 +43,6 @@ int main(int argc, char *argv[])
 
     logFileReader.readAll(filename);
 
-    // Quit the program when it is done.
```
enochtsang: This file didn't need to be edited.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,12 +1,17 @@
...
 {
     const double BATTERY_AMP_HOUR_CAPACITY = 123.0;
+    const double HOURS_TO_MS_CONVERSION = 3.6E6;
+    const double MS_TO_HOURS_CONVERSION = 2.7778E-7;
 }
 
 BatteryStateOfChargeService::BatteryStateOfChargeService(double initialStateOfChargePercent)
-: initialStateOfChargePercent_(initialStateOfChargePercent)
+    : initialStateOfChargePercent_(initialStateOfChargePercent)
+, totalAmpHoursUsed_((100 - initialStateOfChargePercent_) / 100 * BATTERY_AMP_HOUR_CAPACITY)
```
bill-luu: Line up the commas with the semicolon on line 12.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,12 +1,17 @@
...
+, totalCurrent_(0)
```
bill-luu: As above.

Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -1,18 +1,21 @@
...
 
 BatteryStateDisplayService::BatteryStateDisplayService(
     const I_BatteryDataSource& batteryDataSource,
     I_BatteryStateOfChargeService& batteryStateOfChargeService)
 : batteryStateOfChargeService_(batteryStateOfChargeService)
 {
-    // This function is what "connects" the signal to the slot. So whenever the
-    // signals it emitted, the slot will be called and the signal arguements
-    // will be passed into the slot.
+
```
bill-luu: Extra new line.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +21,58 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 bool BatteryStateOfChargeService::isCharging() const
 {
-    return false;
+    if (current_ < 0)
+    {
+       return true;
+    }
+    else
+    {
+        return false;
```
bill-luu: Indentation here is off

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +21,58 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    }
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
-}
+    int ms_remaining;
```
bill-luu: Keep variables in `camelCase` instead of `snake_case`

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +21,58 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    }
+    else
+    {
+      ms_remaining = (BATTERY_AMP_HOUR_CAPACITY - totalAmpHoursUsed()) / averageCurrent_ * HOURS_TO_MS_CONVERSION;
+    }
+
+    return QTime(0, 0, 0, 0).addMSecs(ms_remaining);
 
+}
+/*running average needs to be just the one after it, so I need to write the average as the current + past current divided by 2*/
```
bill-luu: For comments meant to be seen by others, avoid the use of "I". It sounds like a mental note instead of one meant to be read.

I would rewrite this to `The running average of the current should be the (current + pastcurrent)/2`. 
Also, comments here should be more descriptive of the overall function, rather than describing a part of it. Consider putting it right above line 67 instead.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +21,58 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 {
-    Q_UNUSED(batteryData);
-    // This is where you can update your variables
+    if (previousTime_.isNull())
+    {
+        previousTime_ = batteryData.time;
+    }
+    QTime presentTime = batteryData.time;
+    int deltaMseconds;
+    deltaMseconds = abs(presentTime.msecsTo(previousTime_));
```
bill-luu: You can combine the above two lines to be `int deltaMseconds = abs(presentTime.msecsTo(previousTime_));`

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +21,58 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    double deltaHours = deltaMseconds * MS_TO_HOURS_CONVERSION;
+    previousTime_ = batteryData.time;
+
+    if (previousCurrent_ == NULL)
+    {
+        previousCurrent_ = batteryData.current;
+    }
+    current_ = batteryData.current;
+    totalCurrent_ = batteryData.current + previousCurrent_;
```
bill-luu: `totalCurrent_` isn't really being used outside `addData()` anymore, so we don't really need it to be a local variable.
Consider combining this with the next line, so
`averageCurrent_ = (batteryData.current + previousCurrent_) / 2`

Advanced-Recruit-Training/src/main.cpp
bill-luu: Still not entirely the same, missing a space here.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -37,34 +36,46 @@ bool LogFileReader::readAll(const QString& fileName)
...
-    batteryData.current = sections.at(2).toDouble();
+    batteryData.voltage = sections.at(1).toDouble(&voltageOk);
 
-    return true;
+    batteryData.current = sections.at(2).toDouble(&currentOk);
 
+    if (!voltageOk || !currentOk)
+    {
+        return false;
+    }
```
bill-luu: Let's add a new line here (Between 72 and 73). Since typically one wouldn't put a new line between `else` and `else if` statements, this `if` statement can look a bit like an `else if` one.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -37,34 +36,46 @@ bool LogFileReader::readAll(const QString& fileName)
...
         }
     }
 
     return true;
 }
 
-/* File input is a csv file in the format of hh:mm:ss:zzz, voltage, current
- * Need to implement error checking for the correct number of values and
- * that the conversion from string to double is sucessful.*/
+
```
bill-luu: Extra new line here.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
```diff
@@ -1,28 +1,31 @@
...
+#include <QString>
 
 class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
 {
 public:
     explicit BatteryStateOfChargeService(double initialStateOfChargePercent);
     virtual ~BatteryStateOfChargeService();
 
-    /*Returns the total amp hours used*/
+
```
bill-luu: Here, and on lines 15, 18 and 21, there are now extra new lines from where the comment were deleted. Let's get rid of them.

bill-luu:
LGTM! Enoch can take a look at it now.

Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -25,9 +27,20 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
 {
     batteryStateOfChargeService_.addData(batteryData);
 
+    QString TIME_PARAMETER;
```
enochtsang: Only constants should be in CAP_CASE, this is a normal variable and should be in camelCase. Also maybe try to think of a better name, it's not really a parameter and it doesn't really indicate time either.

Also for safety, you should initialize the variable. It helps the longevity of the code.
```c++
QString TIME_PARAMETER = " Time remaining until Battery is Depleted: ";
     if (batteryStateOfChargeService_.isCharging())
     {
         TIME_PARAMETER = " Time remaining until Battery is Charged: ";
     }
```

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +20,55 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 {
-    Q_UNUSED(batteryData);
-    // This is where you can update your variables
+    if (previousTime_.isNull())
+    {
+        previousTime_ = batteryData.time;
+    }
+    QTime presentTime = batteryData.time;
+    int deltaMseconds = abs(presentTime.msecsTo(previousTime_));
+    double deltaHours = deltaMseconds * MS_TO_HOURS_CONVERSION;
```
enochtsang: You can run into type issues here. 

See the example below.
```
double = int * double
double = int
double = int // casted as double
```
With actual numbers this would look like:
```
x = 2 * 2.2
x = 4 // answer is 4.4 but gets casted to int
x = 4.0 // The int gets converted to a double
```

The solution would be to cast the int to double before multiplying.
```
double deltaHours = (double)deltaMseconds * MS_TO_HOURS_CONVERSION;
```

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +20,55 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    if (previousCurrent_ == NULL)
+    {
+        previousCurrent_ = batteryData.current;
+    }
+    current_ = batteryData.current;
+    averageCurrent_ = (batteryData.current + previousCurrent_) / 2;
+    double deltaAh;
+    deltaAh =  averageCurrent_ * deltaHours;
+    totalAmpHoursUsed_ += deltaAh;
+    previousCurrent_ = batteryData.current;
```
enochtsang: Deal with the currents all at once and move this up.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,21 +20,53 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    previousTime_ = batteryData.time;
+
+    if (previousCurrent_ == NULL)
+    {
+        previousCurrent_ = batteryData.current;
+    }
+    current_ = batteryData.current;
+    averageCurrent_ = (batteryData.current + previousCurrent_) / 2;
+    totalAmpHoursUsed_ += averageCurrent_ * deltaHours;
+    previousCurrent_ = batteryData.current;
```
enochtsang: By the end of this function. previousCurrent_ and current_ have the same value and so it's pointless to keep both, what do you think the proper fix for this?

enochtsang:
Good job, finished.

Issue #42, opened 2017-06-17
Add libraries and dependencies to install.sh


bill-luu:
As I was setting up the embedded systems environment, I found that I needed to install the following extra libraries to get it to work:

32 Bit Libraries

- `lib32ncurses5` 
- `libncurses5:i386`
Other
- `software-properties-common`

Let's add this in to the `Recruit-Resources/STM32F4Discovery-Resources/install.sh`

bill-luu:
Also, let's change the install script so that it uses the team's forked versions of stlink, CubeMX2MakeFile, and gcc4mbed, in case any updates occur and mess us up.

bill-luu:
Links to STLink and gcc4mbed related issues:
https://github.com/UCSolarCarTeam/stlink/issues/1
https://github.com/UCSolarCarTeam/gcc4mbed/issues/1

PR #48, opened 2017-08-20
Added libraries and dependencies

jeffh1996:
Regarding issue #42 . I have added the additional libraries lib32ncurses5, libncurses5:i386, and software-properties-common.

Also, the stlink, CubeMX2MakeFile, and gcc4mbed versions now come from the team's repository.

### DIFF
--- a/STM32F4Discovery-Resources/install.sh
+++ b/STM32F4Discovery-Resources/install.sh
@@ -11,7 +11,7 @@ apt-get -y install build-essential git libsane:i386 ia32-libs-multiarch autoconf
 # Install STLink
 if ! type "st-flash" > /dev/null; then
     echo "*** Installing ST Link"
-    git clone https://github.com/texane/stlink.git /opt/stlink
+    git clone https://github.com/UCSolarCarTeam/stlink.git /opt/stlink
     mkdir /opt/stlink/build
     (cd /opt/stlink/build && cmake -DCMAKE_BUILD_TYPE=Debug .. && make -j4)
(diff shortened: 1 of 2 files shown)


enochtsang:
@jeffh1996 reference the issue number next time

bill-luu:
Thanks Jeff! I assigned you issues on the stlink and gcc4mbed repos, can you take care of those too and then we'll test this script to see if it works.

STM32F4Discovery-Resources/install.sh
```diff
@@ -11,7 +11,7 @@ apt-get -y install build-essential git libsane:i386 ia32-libs-multiarch autoconf
 # Install STLink
 if ! type "st-flash" > /dev/null; then
     echo "*** Installing ST Link"
-    git clone https://github.com/UCSolarCarTeam/stlink.git /opt/stlink --depth 1
```
bill-luu: How come this was changed back to the `texane` repo instead of the `UCSolarCarTeam` one?

jeffh1996:
Sorry about that. Fixed that line. Also, I have no idea why travis is failing.

bill-luu:
Having Louis test this

louisj381:
Through testing, found that script ran properly, however when gcc-arm-none-eabi is added to .profile, it may not be updating through path. When running the make command I was getting an error that "gcc-arm... is not found". When I tried another terminal I had the same error. However, when I restarted my computer and ran the command it was successful.

PR #66, opened 2017-11-18
Justin/advanced training

chordatgh1234:


Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,15 +57,31 @@ bool LogFileReader::readAll(const QString& fileName)
  * that the conversion from string to double is sucessful.*/
 bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) const
 {
+    bool check = true;
```
louisj381: I like the optimisation of your code, however it could be more readable and explicit about whats happening, instead of using 'check' for both the voltage and the current, how could you use an extra variable and do a final check at the end?

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,15 +57,31 @@ bool LogFileReader::readAll(const QString& fileName)
...
+
     QStringList sections = line.split(BATDATA_DELIMITER);
 
     QString timeString = sections.at(0);
+
+    if (sections.size() != 3)
```
louisj381: Lets try to avoid 'magic numbers' as much as possible, better to declare a constant in the namespace above.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,17 +26,20 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    return totalAmpHoursUsed_;
 }
 
 bool BatteryStateOfChargeService::isCharging() const
 {
-    return false;
+   if (batteryData_.current < 0)
+       return true;
+   else
+       return false;
```
louisj381: Still follow style guidelines and have braces, even if they are one-line bodies for the if/else statements.

Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -27,7 +27,13 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
 
     QTextStream(stdout) << "Voltage: " << batteryData.voltage
         << " Current: " << batteryData.current
-        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed() << endl;
+        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed();
+
+    QTime timeWhenChargedOrDeplete = batteryStateOfChargeService_.timeWhenChargedOrDepleted();
```
louisj381: lets fix up this variable name :)

Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -27,7 +27,13 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
...
+
+    if(batteryStateOfChargeService_.isCharging())
+        QTextStream(stdout) << " Time till charged: " << timeWhenChargedOrDeplete.toString("hh:mm:ss.zzz") << endl;
+    else
+        QTextStream(stdout) << " Time till depleted: " << timeWhenChargedOrDeplete.toString("hh:mm:ss.zzz") << endl;
```
louisj381: same as above, string could be a bit more formal

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -8,6 +11,13 @@ namespace
 BatteryStateOfChargeService::BatteryStateOfChargeService(double initialStateOfChargePercent)
 : initialStateOfChargePercent_(initialStateOfChargePercent)
 {
+    batteryData_.current = 0;
+    batteryData_.time.setHMS(0,0,0,0);
+    batteryData_.voltage = 0;
+    totalAmpHoursUsed_ = 0;
+    ampHours_ = (initialStateOfChargePercent_/100)*BATTERY_AMP_HOUR_CAPACITY;
+    dataPoints_ = 0;
+    timeDiff_ = 0;
```
louisj381: are these private variable defined in the header files? Also, ampHours_ doesn't quite seem to be logically correct, think about what it is actually means by the initalStateOfCharge.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -35,4 +48,36 @@ void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
     // This is where you can update your variables
     // Hint: There are many different ways that the totalAmpHoursUsed can be updated
     // i.e: Taking a running average of your data values, using most recent data points, etc.
+
+    // The section below is updates the total Amphours used
+    if (dataPoints_ > 0)
+        timeDiff_ = (double)(batteryData_.time.msecsTo(batteryData.time)*0.000000278);
+
```
louisj381: fix up those magic numbers, see commet below

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -8,6 +11,13 @@ namespace
 BatteryStateOfChargeService::BatteryStateOfChargeService(double initialStateOfChargePercent)
 : initialStateOfChargePercent_(initialStateOfChargePercent)
 {
+    batteryData_.current = 0;
+    batteryData_.time.setHMS(0,0,0,0);
+    batteryData_.voltage = 0;
+    totalAmpHoursUsed_ = 0;
+    ampHours_ = (initialStateOfChargePercent_/100)*BATTERY_AMP_HOUR_CAPACITY;
```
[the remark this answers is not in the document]
bill-luu: Also, use initialization lists for initializing private variables.
`BatteryStateOfChargeService::BatteryStateOfChargeService(double initialStateOfChargePercent)`
` : initialStateOfChargePercent_(initialStateOfChargePercent)`
` : ampHours_(initialStateOfChargePercent_/100)*BATTERY_AMP_HOUR_CAPACITY)`
` : etc...`

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -35,4 +48,36 @@ void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
     // This is where you can update your variables
```
bill-luu: These comments can be removed.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -35,4 +48,36 @@ void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
...
     // Hint: There are many different ways that the totalAmpHoursUsed can be updated
     // i.e: Taking a running average of your data values, using most recent data points, etc.
+
+    // The section below is updates the total Amphours used
+    if (dataPoints_ > 0)
```
bill-luu: Since we're only checking if `dataPoints_` is 0 or not, I would suggest to use a boolean instead to indicate if it's the first run.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -35,4 +48,36 @@ void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
...
+        timeDiff_ = (double)(batteryData_.time.msecsTo(batteryData.time)*0.000000278);
+
+    batteryData_.time = batteryData.time;
+
+    if (dataPoints_ == 0)
+        totalAmpHoursUsed_ = ampHours_;
+    if (dataPoints_ > 0)
+        totalAmpHoursUsed_ += (((batteryData.current + batteryData_.current) /2) * timeDiff_);
```
bill-luu: Space between operators.
` totalAmpHoursUsed_ += (((batteryData.current + batteryData_.current) / 2) * timeDiff_);`

bill-luu: Also, this one-liner works, but for simpler readability, I would suggest that we split up the calculation.

bill-luu: Lastly, this `if` statement should be an `else if`, and can be combined with the other one before it.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -35,4 +48,36 @@ void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
...
+
+    batteryData_.current = batteryData.current;
+
+    dataPoints_++;
+
+    // The section below is for calculating the time it takes to charge or deplete
+    double mSeconds;
+    timeWhenChargedOrDepleted_.setHMS(0,0,0,0);
+
+    if (batteryData_.current < 0)
```
bill-luu: Here, we're checking if the battery is charging, so let's use `isCharging()`!

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -35,4 +48,36 @@ void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
...
+    {
+       mSeconds = (totalAmpHoursUsed_/(batteryData_.current*(-1)))*3600000;
```
bill-luu: Instead of multiplying by -1, use the absolute value.
Also, the constant `3600000` is what we would call a "Magic Number" as in a number that just kinda shows up with no real explanation. For these numbers, put them in the namespace as a const int, just like in logfilereader.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -35,4 +48,36 @@ void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
...
+        totalAmpHoursUsed_ = ampHours_;
+    if (dataPoints_ > 0)
+        totalAmpHoursUsed_ += (((batteryData.current + batteryData_.current) /2) * timeDiff_);
+
+    batteryData_.current = batteryData.current;
+
+    dataPoints_++;
+
+    // The section below is for calculating the time it takes to charge or deplete
+    double mSeconds;
```
bill-luu: This can be an `int`, then typecasting won't be needed later.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,15 +57,31 @@ bool LogFileReader::readAll(const QString& fileName)
...
+        return false;
+        //QTextStream(stdout) <<"error ";
+
     batteryData.time = QTime::fromString(timeString, STRING_TIME_FORMAT);
 
-    batteryData.voltage = sections.at(1).toDouble();
+    batteryData.voltage = sections.at(1).toDouble(&check);
 
-    batteryData.current = sections.at(2).toDouble();
+    batteryData.current = sections.at(2).toDouble(&check);
```
bill-luu: There's a bit of a logical error here. If the voltage has an error, but the current is fine then `check` would still return true... how can we fix that up?

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,15 +57,31 @@ bool LogFileReader::readAll(const QString& fileName)
...
 
-    return true;
 
+    if (!batteryData.time.isValid())
+        return false;
+        // QTextStream(stdout) <<"error ";
```
bill-luu: Remove this bit of commented out code.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,15 +57,31 @@ bool LogFileReader::readAll(const QString& fileName)
...
+
+    return check;
 }
+
```
bill-luu: Lots of extra whitespace here!

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -35,4 +48,36 @@ void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
...
     // i.e: Taking a running average of your data values, using most recent data points, etc.
+
+    // The section below is updates the total Amphours used
+    if (dataPoints_ > 0)
+        timeDiff_ = (double)(batteryData_.time.msecsTo(batteryData.time)*0.000000278);
+
+    batteryData_.time = batteryData.time;
+
+    if (dataPoints_ == 0)
+        totalAmpHoursUsed_ = ampHours_;
```
bill-luu: Also, it looks like `ampHours_` is only used to set up `totalAmpHoursUsed_`, I would suggest not having this as a private variable.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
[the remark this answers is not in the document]
bill-luu: Lastly, this would set `ampHours_` to be the amphours in the battery. However you set the amp hours _used_ as this value.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +25,62 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+
+    if (checkStart_ == true)
+    {
+        totalAmpHoursUsed_ = (initialStateOfChargePercent_ / 100) * BATTERY_AMP_HOUR_CAPACITY;
+    }
+
+    else if (checkStart_ == false)
+    {
+        timeDiff_ = (double)(time_.msecsTo(batteryData.time) * MSECONDS_TO_HOURS_CONVERSION);
```
bill-luu: It's a bit cleaner if we did
`timeDiff_ = (double)(time_.msecsTo(batteryData.time)) * MSECONDS_TO_HOURS_CONVERSION;`

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +25,62 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+        double currentAverage = (batteryData.current + current_) / 2;
+        totalAmpHoursUsed_ += currentAverage * timeDiff_;
+    }
+
+    time_ = batteryData.time;
+
+    current_ = batteryData.current;
+
+    checkStart_ = false;
```
bill-luu: This only needs to be set if checkStart_ was true, so we can place that in the if above.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
```diff
@@ -25,4 +26,10 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     const double initialStateOfChargePercent_;
 
     // Add your necessary variables here
+    double current_;
+    double totalAmpHoursUsed_;
+    bool checkStart_;
+    double timeDiff_;
+    QTime time_;
```
bill-luu: Try and be a bit more specific with these naming, ie is this the time of the "Present time" data point, or the time of the "Previous" data point? Same goes for the current.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,15 +57,34 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    batteryData.current = sections.at(2).toDouble(&currentCheck);
 
-    return true;
 
+    if (!batteryData.time.isValid())
+    {
+        return false;
+    }
+
+    if (voltageCheck == true && currentCheck == true )
```
bill-luu: This if statement can be combined with the previous one. Also, use the `if(boolean)` format for better readability.

PR #67, opened 2017-11-18
Magnus/advanced training

mglyngberg:
I completed all of the tasks for advanced recruit training.

### DIFF
--- a/Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
+++ b/Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
@@ -27,7 +27,17 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
 
     QTextStream(stdout) << "Voltage: " << batteryData.voltage
         << " Current: " << batteryData.current
-        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed() << endl;
+        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed();
+    if(batteryStateOfChargeService_.isCharging())
+    {
+        QTextStream(stdout) << "    Time Until Charged: " << batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString("hh:mm:ss.zzz")
+                            << endl;
+    }
+    else
+    {
+        QTextStream(stdout) << "    Time Until Depletion: " << batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString("hh:mm:ss.zzz")
+                            << endl;
+    }
 
     // TODO: Print out time until the battery is fully charged or depleted.
 }
--- a/Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
+++ b/Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
@@ -38,6 +38,4 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
         QTextStream(stdout) << "    Time Until Depletion: " << batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString("hh:mm:ss.zzz")
                             << endl;
     }
-
-    // TODO: Print out time until the battery is fully charged or depleted.
 }
--- a/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
+++ b/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
@@ -3,16 +3,14 @@
 namespace
 {
     const double BATTERY_AMP_HOUR_CAPACITY = 123.0;
+    const double MILLISECONDS_IN_HOUR = 3600000;
 }
 
 BatteryStateOfChargeService::BatteryStateOfChargeService(double initialStateOfChargePercent)
 : initialStateOfChargePercent_(initialStateOfChargePercent)
+, current_(0)
+, ampHoursUsed_ (BATTERY_AMP_HOUR_CAPACITY * (1 - (initialStateOfChargePercent_ / 100)))
 {
-    current_ = 0;
-    ampHoursUsed_ = BATTERY_AMP_HOUR_CAPACITY - BATTERY_AMP_HOUR_CAPACITY*(initialStateOfChargePercent_/100);
-    time_.setHMS(0,0,0,0);
-    chargeTime_.setHMS(0,0,0,0);
-    numberOfRecords_ = 0;
 }
 
 BatteryStateOfChargeService::~BatteryStateOfChargeService()
@@ -27,37 +25,48 @@ double BatteryStateOfChargeService::totalAmpHoursUsed() const
 bool BatteryStateOfChargeService::isCharging() const
 {
     if (current_ < 0 )
+    {
         return true;
+    }
     else
+    {
         return false;
+    }
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return chargeTime_;
+    double hours;
+    QTime timeToChargeOrDepletion(0,0,0,0);
+
+    if (isCharging())
+    {
+        hours = ampHoursUsed_ / qAbs(current_);
+    }
+    else
+    {
+        hours = (BATTERY_AMP_HOUR_CAPACITY - ampHoursUsed_) / current_;
+    }
+
+    timeToChargeOrDepletion = timeToChargeOrDepletion.addMSecs(hours * MILLISECONDS_IN_HOUR);
+    return timeToChargeOrDepletion;
 }
 
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    double timeDifference;
-    double hours;
-    chargeTime_.setHMS(0,0,0,0);
+    double timeDifference = 0;
 
-    if (numberOfRecords_ > 0)
+    if (isFirstRecord_)
     {
-        timeDifference = (double)time_.msecsTo(batteryData.time)/3600000;
+        isFirstRecord_ = false;
     }
-    numberOfRecords_++;
+    else
+    {
+        timeDifference = (double)time_.msecsTo(batteryData.time) / MILLISECONDS_IN_HOUR;
+    }
 
-    ampHoursUsed_ += ((current_ + batteryData.current)/2)*timeDifference;
+    ampHoursUsed_ += ((current_ + batteryData.current) / 2) * timeDifference;
 
     time_ = batteryData.time;
     current_ = batteryData.current;
-
-    if (isCharging())
-         hours = -ampHoursUsed_/current_;
-    else
-        hours = (BATTERY_AMP_HOUR_CAPACITY - ampHoursUsed_)/current_;
-
-    chargeTime_ = chargeTime_.addMSecs(hours*3600000);
 }
--- a/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
+++ b/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
@@ -8,6 +8,11 @@ namespace
 BatteryStateOfChargeService::BatteryStateOfChargeService(double initialStateOfChargePercent)
 : initialStateOfChargePercent_(initialStateOfChargePercent)
 {
+    current_ = 0;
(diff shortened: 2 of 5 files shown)


Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -8,6 +8,11 @@ namespace
 BatteryStateOfChargeService::BatteryStateOfChargeService(double initialStateOfChargePercent)
 : initialStateOfChargePercent_(initialStateOfChargePercent)
 {
+    current_ = 0;
+    ampHoursUsed_ = BATTERY_AMP_HOUR_CAPACITY - BATTERY_AMP_HOUR_CAPACITY*(initialStateOfChargePercent_/100);
```
louisj381: while this appears to be logically correct, perhaps there is a more straightforward way to accomplish this calculation, say without reusing the same variable twice?

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +21,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    double hours;
+    chargeTime_.setHMS(0,0,0,0);
+
+    if (numberOfRecords_ > 0)
+    {
+        timeDifference = (double)time_.msecsTo(batteryData.time)/3600000;
+    }
+    numberOfRecords_++;
+
+    ampHoursUsed_ += ((current_ + batteryData.current)/2)*timeDifference;
```
louisj381: what happens the first time it runs through, if numberOfRecords_ == 0? won't timeDifference be undefined?

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +21,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
-    // This is where you can update your variables
-    // Hint: There are many different ways that the totalAmpHoursUsed can be updated
-    // i.e: Taking a running average of your data values, using most recent data points, etc.
+    double timeDifference;
+    double hours;
+    chargeTime_.setHMS(0,0,0,0);
+
+    if (numberOfRecords_ > 0)
+    {
+        timeDifference = (double)time_.msecsTo(batteryData.time)/3600000;
```
louisj381: this conversion uses a magic number, eliminate those by utilising the namespace above :)

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,14 +57,26 @@ bool LogFileReader::readAll(const QString& fileName)
  * that the conversion from string to double is sucessful.*/
 bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) const
 {
+    bool valid =true;
+
     QStringList sections = line.split(BATDATA_DELIMITER);
 
     QString timeString = sections.at(0);
+
+    if (sections.size() !=3)
```
louisj381: another magical number (3)

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,14 +57,26 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    if(!batteryData.time.isValid())
+        return false;
+
+    batteryData.voltage = sections.at(1).toDouble(&valid);
 
-    batteryData.current = sections.at(2).toDouble();
+    batteryData.current = sections.at(2).toDouble(&valid);
+
+    if (!valid)
+        return false;
```
louisj381: follow style guildlines and show me those nice brackets when doing if/else statements. Also, recommended you don't use one boolean to check to different sections, while it may be more efficient, it isn't as clear

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +21,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+    time_ = batteryData.time;
+    current_ = batteryData.current;
+
+    if (isCharging())
+         hours = -ampHoursUsed_/current_;
+    else
+        hours = (BATTERY_AMP_HOUR_CAPACITY - ampHoursUsed_)/current_;
+
+    chargeTime_ = chargeTime_.addMSecs(hours*3600000);
```
louisj381: it seems that some of this would fit better in the TimeWhenChargedOrDepleted function.

Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -27,7 +27,17 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
...
+        QTextStream(stdout) << "    Time Until Charged: " << batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString("hh:mm:ss.zzz")
+                            << endl;
+    }
+    else
+    {
+        QTextStream(stdout) << "    Time Until Depletion: " << batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString("hh:mm:ss.zzz")
+                            << endl;
+    }
 
     // TODO: Print out time until the battery is fully charged or depleted.
```
bill-luu: Remove this TODO

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -8,6 +8,11 @@ namespace
 BatteryStateOfChargeService::BatteryStateOfChargeService(double initialStateOfChargePercent)
 : initialStateOfChargePercent_(initialStateOfChargePercent)
 {
+    current_ = 0;
+    ampHoursUsed_ = BATTERY_AMP_HOUR_CAPACITY - BATTERY_AMP_HOUR_CAPACITY*(initialStateOfChargePercent_/100);
+    time_.setHMS(0,0,0,0);
```
bill-luu: Use initialization lists for initializing private variables.
`BatteryStateOfChargeService::BatteryStateOfChargeService(double initialStateOfChargePercent)`
`: initialStateOfChargePercent_(initialStateOfChargePercent)`
`: ampHoursUsed_BATTERY_AMP_HOUR_CAPACITY - BATTERY_AMP_HOUR_CAPACITY*(initialStateOfChargePercent_/100);`
`: etc...`

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +21,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    if (current_ < 0 )
+        return true;
+    else
+        return false;
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    return chargeTime_;
```
bill-luu: This variable could technically be the depletion time as well, so let's fix up the name

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
bill-luu: Also, let's keep things consistent by adding a space between operators.
`timeDifference = (double)time_.msecsTo(batteryData.time) / 3600000;`

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
bill-luu: Space between operators

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +21,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    }
+    numberOfRecords_++;
+
+    ampHoursUsed_ += ((current_ + batteryData.current)/2)*timeDifference;
+
+    time_ = batteryData.time;
+    current_ = batteryData.current;
+
+    if (isCharging())
+         hours = -ampHoursUsed_/current_;
```
bill-luu: Instead of applying a negative, use the absolute value.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +21,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 }
 
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
-    // This is where you can update your variables
-    // Hint: There are many different ways that the totalAmpHoursUsed can be updated
-    // i.e: Taking a running average of your data values, using most recent data points, etc.
+    double timeDifference;
+    double hours;
```
bill-luu: This variable isn't used until much later, group them up so they would look like a singular "Idea"!

Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +21,43 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    chargeTime_.setHMS(0,0,0,0);
+
+    if (numberOfRecords_ > 0)
```
bill-luu: Since we're only using `numberOfRecords_` to see if we have more than one record, I would suggest using a boolean to check that.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,14 +57,26 @@ bool LogFileReader::readAll(const QString& fileName)
...
     batteryData.time = QTime::fromString(timeString, STRING_TIME_FORMAT);
 
-    batteryData.voltage = sections.at(1).toDouble();
+    if(!batteryData.time.isValid())
+        return false;
+
+    batteryData.voltage = sections.at(1).toDouble(&valid);
 
-    batteryData.current = sections.at(2).toDouble();
+    batteryData.current = sections.at(2).toDouble(&valid);
```
bill-luu: So if voltage was invalid, but current was fine, what would be the value in `valid`?

Advanced-Recruit-Training/src/LogFileReader.cpp
bill-luu: Adding on to that, reverse the statement, so `if(value) .... else .....` it isn't a big thing, but the reason for this is that the reader wouldn't have to do mental logic to do read statement. That's usually why you'll see booleans be like `isComplete` or `isValid`, it reduces mental overhead and makes larger ifs much easier to read.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,27 +57,31 @@ bool LogFileReader::readAll(const QString& fileName)
  * that the conversion from string to double is sucessful.*/
 bool LogFileReader::parseLine(const QString& line, BatteryData& batteryData) const
 {
-    bool valid =true;
+    bool currentIsValid = true;
```
bill-luu: These variables aren't being used until much later, let's declare them  when they are used.

Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,27 +57,31 @@ bool LogFileReader::readAll(const QString& fileName)
...
 
     QString timeString = sections.at(0);
 
-    if (sections.size() !=3)
+    if (sections.size() != COLUMNS)
         return false;
 
     batteryData.time = QTime::fromString(timeString, STRING_TIME_FORMAT);
 
     if(!batteryData.time.isValid())
```
bill-luu: Combine this with the other isValid ifs.

Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
```diff
@@ -28,6 +28,5 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     double current_;
     double ampHoursUsed_;
     QTime time_;
```
bill-luu: Try and be a bit more specific with these naming, ie is this the time of the "Present time" data point, or the time of the "Previous" data point? Same goes for the current.

PR #135, opened 2019-12-05
Jessie Advance recruit training

JessieG123:


Recruit-Training/Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -25,9 +26,29 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
 {
     batteryStateOfChargeService_.addData(batteryData);
 
+    double milisec = batteryStateOfChargeService_.timeWhenChargedOrDepleted().msec();
```
navjotbrar: millisec*, want to keep variable names clear.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -25,9 +26,29 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
...
+    if(milisec < 10){
+        addZero = "00";
+    }
+
     QTextStream(stdout) << "Voltage: " << batteryData.voltage
         << " Current: " << batteryData.current
-        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed() << endl;
+        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed();
+    if(batteryStateOfChargeService_.isCharging()){
+        QTextStream(stdout) << " Time until charged: " << batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString() <<"." << addZero << milisec <<endl;
```
navjotbrar: When lines are too long, just place some of the line below it with the ENTER key.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -25,9 +26,29 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
...
+
+    }else{
 
+        QTextStream(stdout) << " Time until depleted: " << batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString() << "."
+                     << addZero <<milisec <<endl;
+    }
     // TODO: Print out time until the battery is fully charged or depleted.
```
navjotbrar: Remove any unnecessary comments

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +22,72 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+    return totalAmphoursUsed_;
 }
 
 bool BatteryStateOfChargeService::isCharging() const
 {
+    if(current_ < 0){
+        return true;
+    }
     return false;
```
navjotbrar: Can use one return statement here to encompass all of this code. Ex: return (current<0)

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +22,72 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    double totalHoursRemaining, msRemaining;
+        int h, m, s, ms;
```
navjotbrar: Use more descriptive variable names to keep code clean

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +22,72 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+        if(isCharging())
+            totalHoursRemaining = totalAmpHoursUsed() / current_ * -1;
+        else
+            totalHoursRemaining = (BATTERY_AMP_HOUR_CAPACITY - totalAmpHoursUsed()) / current_;
+
+        while(totalHoursRemaining > 24){
+            totalHoursRemaining -= 24;
+        }
+        //Converting totalHoursRemaining into h, m, s, ms
```
navjotbrar: Just a reminder to remove unnecessary comments, want to make code that is self-documenting.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +22,72 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+        h = totalHoursRemaining;
+
+        msRemaining = (totalHoursRemaining - h) * 3600000;
+        m = msRemaining / 60000;
+        msRemaining = msRemaining - (m * 60000);
+        s = msRemaining / 1000;
+        msRemaining = msRemaining - (s * 1000);
+        ms = msRemaining;
```
navjotbrar: Just a rule of thumb, instead of using numbers such as 3600000 in your code, you want to place these numbers in a variable or macro to make code readable. If code has numbers without context or are not easily readable, we call them magic numbers

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.h
```diff
@@ -29,4 +29,10 @@ class BatteryStateOfChargeService : public I_BatteryStateOfChargeService
     const double initialStateOfChargePercent_;
 
     // Add your necessary variables here
+    double current_;
+    double oldCurrent_;
+    QTime newTime_;
+    QTime oldTime_;
+    double currentAmphoursUsed_ = 0;
```
navjotbrar: Be consistent when initializing your variables, usually initialize within the constructor. currentAmpHoursUsed_ should be initialized withing the constructor

Recruit-Training/Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,15 +58,30 @@ bool LogFileReader::readAll(const QString& fileName)
...
-    batteryData.voltage = sections.at(1).toDouble();
 
-    batteryData.current = sections.at(2).toDouble();
 
-    return true;
+    QTime time1 = QTime::fromString(timeString, STRING_TIME_FORMAT);
+    batteryData.time = time1;
+    bool isTimeValid = QTime::isValid(time1.hour(), time1.minute(), (time1.second() + (time1.msec()/1000)));
+    bool flag1;
+    bool flag2;
```
navjotbrar: Initialize the variables to show there explicit beginning state. Also use descriptive names

Recruit-Training/Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,15 +58,30 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    batteryData.voltage = sections.at(1).toDouble(&flag1);
+    batteryData.current = sections.at(2).toDouble(&flag2);
+    if(flag1 && flag2 && isTimeValid){
+        return true;
+    }else{
+        return false;
+    }
```
navjotbrar: As practice we usually write brackets on newlines. This also can be encompassed through one return statement.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,8 +1,14 @@
 #include "BatteryStateOfChargeService.h"
+#include "BatteryData.cpp"
```
bill-luu: Including a source file inside another source file is a **VERY bad** thing. It can lead to `multiple definition` errors, which is actually what I got when switching between branches.

What ends up happening is C++ will try to compile `BatteryStateOfChargeService.cpp`, and since it includes `BatteryData.cpp`, then that obj file will also have symbols from `BatteryData.cpp`. Since .cpps don't have include guards, it will then try to compile `BatteryData.cpp`, and it'll be super confused since the symbols are already defined.

I think it worked for you since QT must be smart enough to realise when it's already built a symbol, but it's not perfect and will cause lots of cancer down the line if the compiler changes or a project gets bigger. 

In other projects, you might see people using c++ templates by including a `.cpp` somewhere, but that's still pretty cancerous, and they should be using `_impl.h` instead.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -25,9 +26,29 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
 {
     batteryStateOfChargeService_.addData(batteryData);
 
+    double milisec = batteryStateOfChargeService_.timeWhenChargedOrDepleted().msec();
+    QString addZero;
+    if(milisec < 100){
+        addZero = "0";
```
bill-luu: Instead of padding it like this, try formatting your output instead:
https://stackoverflow.com/questions/2618414/convert-an-int-to-a-qstring-with-zero-padding-leading-zeroes

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -25,9 +26,29 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
...
+    }
+
     QTextStream(stdout) << "Voltage: " << batteryData.voltage
         << " Current: " << batteryData.current
-        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed() << endl;
+        << " Total Ah used: " << batteryStateOfChargeService_.totalAmpHoursUsed();
+    if(batteryStateOfChargeService_.isCharging()){
+        QTextStream(stdout) << " Time until charged: " << batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString() <<"." << addZero << milisec <<endl;
+
+    }else{
```
bill-luu: General if/else format on the team is:
```
if(blah)
{
//stuff
}
else
{
//stuff
}
```

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -25,9 +26,29 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
...
 
+        QTextStream(stdout) << " Time until depleted: " << batteryStateOfChargeService_.timeWhenChargedOrDepleted().toString() << "."
+                     << addZero <<milisec <<endl;
```
bill-luu: Add spaces between operators, 
`<< addZero <<milisec <<endl;` -> `<< addZero << milisec << endl;`

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,8 +1,14 @@
...
+#include <QTime>
+#include <QLocale>
+#include <cmath>
+#include <QTextStream>
```
bill-luu: It doesn't look like any of these includes are being used.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,8 +1,14 @@
...
 
 namespace
 {
     const double BATTERY_AMP_HOUR_CAPACITY = 123.0;
+
```
bill-luu: Excess whitespace

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +22,72 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
 
 double BatteryStateOfChargeService::totalAmpHoursUsed() const
 {
-    return 0.0;
+
```
bill-luu: Excess whitespace

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
bill-luu: Also, we should separate variable declarations to one per line.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +22,72 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
     return false;
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+    double totalHoursRemaining, msRemaining;
+        int h, m, s, ms;
+
+        if(isCharging())
```
bill-luu: In general, even if what follows is just one line, we should still encapsulate the statement in brackets. That way in the future, people can add stuff to the statement as needed without accidentally causing bugs.

image

^ Apple's goto fail is an example of that (also never use goto!)

Propagate this to your other if/else statements and loops

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
bill-luu: https://en.wikipedia.org/wiki/Magic_number_(programming)

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +22,72 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+        QTime timeRemaining(h, m, s, ms);
+        return timeRemaining;
+
 }
 
 void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
 {
-    Q_UNUSED(batteryData);
+    double initialAmphoursUsed = (initialStateOfChargePercent_ * BATTERY_AMP_HOUR_CAPACITY)/100;
+    Q_UNUSED(batteryData)
```
bill-luu: Not unused anymore, delete this!

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +22,72 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
     // This is where you can update your variables
     // Hint: There are many different ways that the totalAmpHoursUsed can be updated
     // i.e: Taking a running average of your data values, using most recent data points, etc.
+
+    oldCurrent_ = current_;
```
bill-luu: `oldCurrent_` doesn't need to be a private variable, as it's lifespan is just this one function.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +22,72 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    current_ = batteryData.current;
+    oldTime_ = newTime_;
```
bill-luu: `oldTime_` doesn't need to be a private variable, as it's lifespan is just this one function.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +22,72 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    newTime_ = batteryData.time;
+    double averageCurrent = (oldCurrent_ + current_)/2;
+
+    double changeInTimeMseconds = oldTime_.msecsTo(newTime_);
+    //use secsTo to get the number of seconds elasped from old to new?
+    double seconds = changeInTimeMseconds/1000;
+    double hour = seconds/3600;
```
bill-luu: If this is the only use of seconds, can we just do `changeInTimeMseconds/3600000`?

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +22,72 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    currentAmphoursUsed_ = (hour * averageCurrent);
+    totalAmphoursUsed_ = totalAmphoursUsed_ + currentAmphoursUsed_ + initialAmphoursUsed ; //need to include the totalcurrent again and maybe initialize totalAmphoursUsed to be the intitialAmphoursUsed instead ...
```
bill-luu: It looks like you and Duan did the same thing, so I'll copy/paste my comment on his:

From what it looks like, it looks like the steps you're doing is:

1. Calculate the initial amp hours used
2. Get the current amphours used
3. Add those two together to get the totalAmpHours used
4. In the first case, the currentAmphoursUsed will be 0, so totalAmpHoursUsed==InitialAmphoursUsed
5. Otherwise, take the initialAmphoursUsed and thus totalAmpHoursUsed==currentAmpHoursUsed
This is a veeery roundabout way of doing it haha, I think it makes more sense to get rid of the initialAmpHoursUsed variable altogether, and in the constructor you can set `totalAmpHoursUsed=initialStateOfChargePercent_ / 100 * BATTERY_AMP_HOUR_CAPACITY;`, and do your calculation from there. If you need to differentiate the first value, some kind of flag can be used.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +22,72 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+
+
+    if(totalAmphoursUsed_ != initialAmphoursUsed){
+        totalAmphoursUsed_ = totalAmphoursUsed_ - initialAmphoursUsed;
+    }
+
+
+
```
bill-luu: Excess whitespace

Recruit-Training/Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,15 +58,30 @@ bool LogFileReader::readAll(const QString& fileName)
...
     QString timeString = sections.at(0);
-    batteryData.time = QTime::fromString(timeString, STRING_TIME_FORMAT);
 
-    batteryData.voltage = sections.at(1).toDouble();
 
-    batteryData.current = sections.at(2).toDouble();
 
-    return true;
+    QTime time1 = QTime::fromString(timeString, STRING_TIME_FORMAT);
+    batteryData.time = time1;
```
bill-luu: You're not doing anything with time1, so it'll be cleaner to just do `batteryData.time = QTime::fromString(timeString, STRING_TIME_FORMAT);`

Recruit-Training/Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,15 +58,30 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    bool isTimeValid = QTime::isValid(time1.hour(), time1.minute(), (time1.second() + (time1.msec()/1000)));
```
bill-luu: You should be able to do batteryData.time.isValid()

Recruit-Training/Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -57,15 +58,30 @@ bool LogFileReader::readAll(const QString& fileName)
...
+    bool flag2;
+    batteryData.voltage = sections.at(1).toDouble(&flag1);
+    batteryData.current = sections.at(2).toDouble(&flag2);
+    if(flag1 && flag2 && isTimeValid){
+        return true;
+    }else{
+        return false;
+    }
+
+
```
bill-luu: Excess whitespace

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateDisplayService.cpp
```diff
@@ -26,29 +27,22 @@ void BatteryStateDisplayService::handleBatteryDataReceived(const BatteryData& ba
...
-    QString addZero;
-    if(milisec < 100){
-        addZero = "0";
-    }
+    double millisec = batteryStateOfChargeService_.timeWhenChargedOrDepleted().msec();
 
-    if(milisec < 10){
-        addZero = "00";
-    }
+    QString numberMillisec = QString("%1").arg((int)(millisec), 3, 10, QChar('0'));
```
bill-luu: I'm pretty sure `.msec()` already returns a int, so just change the type of `millisec`

bill-luu: Alternatively, passing in `batteryStateOfChargeService_.timeWhenChargedOrDepleted().msec();` without using a variable is also good.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -1,19 +1,20 @@
...
+    const double HOUR_TO_MILLISECONDS = 3600000;
+    const double MINUTE_TO_MILLISECONDS = 60000;
+    const double SECOND_TO_MILLISECONDS = 1000;
 }
 
 BatteryStateOfChargeService::BatteryStateOfChargeService(double initialStateOfChargePercent)
 : initialStateOfChargePercent_(initialStateOfChargePercent)
 {
+    currentAmphoursUsed_ = 0;
+    totalAmphoursUsed_ = (initialStateOfChargePercent_ * BATTERY_AMP_HOUR_CAPACITY)/100;
```
bill-luu: Use initialization list to initialize your variables (see the initialStateOfChargePercent_)

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -22,72 +23,68 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    double totalHoursRemaining, msRemaining;
-        int h, m, s, ms;
+        double totalHoursRemaining;
+        double millisecRemaining;
+        int hour;
+        int min;
+        int sec;
+        int millisec;
```
bill-luu: These aren't used until later, and can be set right away:
`int hour =  totalHoursRemaining;`, etc.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -22,72 +23,68 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
-    totalAmphoursUsed_ = totalAmphoursUsed_ + currentAmphoursUsed_ + initialAmphoursUsed ; //need to include the totalcurrent again and maybe initialize totalAmphoursUsed to be the intitialAmphoursUsed instead ...
-
-
-    if(totalAmphoursUsed_ != initialAmphoursUsed){
-        totalAmphoursUsed_ = totalAmphoursUsed_ - initialAmphoursUsed;
-    }
+    double averageCurrent = (oldCurrent + current_)/2;
 
+    double changeInTimeMseconds = oldTime.msecsTo(newTime_);
+    double hour = changeInTimeMseconds/3600000;
```
bill-luu: Add a space between operators:
`changeInTimeMseconds / 3600000;`

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -11,10 +11,10 @@ namespace
 }
 
 BatteryStateOfChargeService::BatteryStateOfChargeService(double initialStateOfChargePercent)
-: initialStateOfChargePercent_(initialStateOfChargePercent)
+: initialStateOfChargePercent_(initialStateOfChargePercent),
+  currentAmphoursUsed_(0),
+  totalAmphoursUsed_((initialStateOfChargePercent_ * BATTERY_AMP_HOUR_CAPACITY)/100)
```
bill-luu: Spaces between operators.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -82,9 +77,7 @@ void BatteryStateOfChargeService::addData(const BatteryData& batteryData)
     double averageCurrent = (oldCurrent + current_)/2;
```
bill-luu: Space between operators

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +23,61 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
 }
 
 QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
 {
-    return QTime::currentTime();
+        double totalHoursRemaining;
+
+        if(isCharging())
+        {
+            totalHoursRemaining = totalAmpHoursUsed() / current_ * -1;
```
bill-luu: Use the absolute value of current instead.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +23,61 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+        }
+        else
+        {
+            totalHoursRemaining = (BATTERY_AMP_HOUR_CAPACITY - totalAmpHoursUsed()) / current_;
+        }
+
+        while(totalHoursRemaining > 24)
+        {
+            totalHoursRemaining -= 24;
```
bill-luu: What is the point of this check? Seems like magic, might benefit from a comment.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -37,13 +37,14 @@ QTime BatteryStateOfChargeService::timeWhenChargedOrDepleted() const
...
         {
-            totalHoursRemaining = totalAmpHoursUsed() / current_ * -1;
+            totalHoursRemaining = abs(totalAmpHoursUsed() / current_);
         }
         else
         {
             totalHoursRemaining = (BATTERY_AMP_HOUR_CAPACITY - totalAmpHoursUsed()) / current_;
         }
 
+        //If the totalHoursReamining is greater than 24 continue to subtract it otherwise isn't valid to display since its more than a day
```
bill-luu: Misspelled "Remaining" :)

But now I understand that you want to work around the 24 hour limitation of QTime. TBH, I don't think it was intended that this would happen, but I'm not sure about the way it's being addressed here.

Conceptually, there's 2 ways of addressing this issue for hours > 24 that people have been doing, ex 60 hours:
1. Return the number of hours total, so you would display 60:00:00.000
2. Allow it to loop around - you display 12:00:00.000

If you want to do option 1, you would have to store the excess hours in a new variable that is accessed later. All the work in calculating the seconds, hours, and minutes done here would be needed.

Option 2 is fine too, and your implementation works, but it's there's a better way. Look into QTime's `addMSecs()` function, which wraps around as the time passes midnight.

bill-luu: Also, a small critique of your comment - `If the totalHoursReamining is greater than 24 continue to subtract it` can be already derived from reading the first line of the while loop. Unless the code isn't straightforward enough to just read, most comments should be supplementary - explaining the intention behind the change, instead of what the change does.

ie.
```
// QTime does not support hours > 24
while(...) 
```

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +21,48 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
     // i.e: Taking a running average of your data values, using most recent data points, etc.
+
+    double oldCurrent = current_;
+    current_ = batteryData.current;
+    QTime oldTime = newTime_;
+    newTime_ = batteryData.time;
+    double averageCurrent = (oldCurrent + current_) / 2;
+
+    double changeInTimeMseconds = oldTime.msecsTo(newTime_);
+    double hour = changeInTimeMseconds / 3600000;
```
bill-luu: Magic number here.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +21,48 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    currentAmphoursUsed_ = hour * averageCurrent;
```
bill-luu: currentAmphoursUsed_ is only used in the context of this function, doesn't need to be a private variable.

Recruit-Training/Advanced-Recruit-Training/src/LogFileReader.cpp
```diff
@@ -10,6 +10,7 @@ namespace
     const QString STRING_TIME_FORMAT= "hh:mm:ss.zzz";
     const QString BATDATA_DELIMITER= ", ";
     const int COLUMNS = 3;
+
```
bill-luu: Extra whitespace.

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +20,48 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    if(isCharging())
+    {
+        totalHoursRemaining = abs(totalAmpHoursUsed() / current_);
+    }
+    else
+    {
+        totalHoursRemaining = (BATTERY_AMP_HOUR_CAPACITY - totalAmpHoursUsed()) / current_;
+    }
+
+    QTime time(0, 0, 0, 0);
```
bill-luu: You can make this `timeRemaining`, and then reuse it later with `timeRemaining = timeRemaining.addMSecs(..)`

Recruit-Training/Advanced-Recruit-Training/src/BatteryStateOfChargeService.cpp
```diff
@@ -16,23 +20,48 @@ BatteryStateOfChargeService::~BatteryStateOfChargeService()
...
+    int millisec =  totalHoursRemaining * HOUR_TO_MILLISECONDS;
```
bill-luu: `totalMillisecsRemaining` would be a better name.

PR #152, opened 2020-09-12
User/islamrahat735/embedded training

islamrahat735:


Recruit-Training/Embedded-Recruit-Training/Core/Src/main.c
```diff
@@ -288,6 +328,41 @@ void HAL_CAN_RxFifo0MsgPendingCallback(CAN_HandleTypeDef* hcan)
     }
 
     //TODO: Match StdId of header and data length content for green and blue messages and check data for set bit of toggling green and blue led
+    if ((hdr.StdId == BLUE_MESSAGE_STDID || hdr.StdId == GREEN_MESSAGE_STDID) && hdr.DLC == 1)
+    {
+        if (data[0] == 1 && data[3] == 1 && data[7] == 1)
```
JessieG123, line 333: I think we mentioned last time that this should be a bitwise operation to check which bits has been set.
To check if bits 0, 3 and 7 is set you can set a bitmask like `10001001`. That way when it's bitwise AND ( `&`) together only bits 0, 3 and 7 are going to be a 1 if they were a 1 before. Since `0 & 0 = 0`, `1 & 0 = 1`, `1 & 1 = 1`.

Recruit-Training/Embedded-Recruit-Training/Core/Src/main.c
```diff
@@ -288,6 +328,41 @@ void HAL_CAN_RxFifo0MsgPendingCallback(CAN_HandleTypeDef* hcan)
...
+        {
+            if (BlueToggleFlag)
+                BlueToggleFlag = 0;
+            else
+                BlueToggleFlag = 1;
+
+        }
+    
+        if (data[0] == 1 && data[1] == 1)
```
JessieG123, line 342: Same thing down here. You can probably get rid of whatever you're doing with a for loop down there when you implement the bitmask correctly. 

Each value of data is 8 bits since it's `uint8_t`. So `data[0]`'s value is 8 bits long. I think you're thinking like if I have a message of `0101` it becomes `data = {0, 1, 0, 1}`, but it's actually like `data={0000 0101, blah, blah}`. 

Correct me if I'm wrong...

Recruit-Training/Embedded-Recruit-Training/Core/Src/BlueLedToggleTask.c
```diff
@@ -11,7 +13,32 @@ void blueLedToggleTask(void const* arg)
...
+        }
+        
+        uint8_t ledState = HAL_GPIO_ReadPin( LED_RED_GPIO_Port, LED_RED_Pin);
         //TODO: Send CAN message indicating current state of LED
+
+        if (osMutexAcquire(canMutex, 0) == osOK)
+        {
+            if (HAL_CAN_GetTxMailboxesFreeLevel(&hcan2) > 0)
+            {
+            uint8_t x[1];
```
JessieG123, line 30: You should probably make the variable more descriptive instead of a `x`

Recruit-Training/Embedded-Recruit-Training/Core/Src/GreenLedToggleTask.c
```diff
@@ -11,7 +17,28 @@ void greenLedToggleTask(void const* arg)
...
+        }
+
+        uint8_t ledState = HAL_GPIO_ReadPin(LED_GREEN_GPIO_Port, LED_GREEN_Pin);
         //TODO: Send CAN message indicating current state of LED
+
+        if (osMutexAcquire(canMutex, 0) == osOK)
+        {
+            if (HAL_CAN_GetTxMailboxesFreeLevel(&hcan2) > 0)
+            {
+            uint8_t x[1];
```
JessieG123, line 34: Same thing here. Make it more descriptive than just a `x` please.

JessieG123 (changes requested):

Recruit-Training/Embedded-Recruit-Training/Core/Src/GreenLedToggleTask.c
```diff
@@ -2,7 +2,13 @@
 
 void greenLedToggleTask(void const* arg)
 {
+    extern CAN_HandleTypeDef hcan2;
+    extern CAN_TxHeaderTypeDef CAN_TX;
+    extern uint8_t GreenToggleFlag;
+    
     //One time osDelayUntil initialization
+    static const uint32_t GREEN_LED_STATUS_STDID = 0xDDD;
+    static const uint32_t GREEN_LED_TOGGLE_FREQ = 10000;
```
JessieG123, line 11: This number is wrong. It should be 10Hz. Use the formula `f=1/t`. Note that 1Hz is 1 second per cycle and we want it in milliseconds.

Recruit-Training/Embedded-Recruit-Training/Core/Src/BlueLedToggleTask.c
```diff
@@ -3,6 +3,8 @@
 void blueLedToggleTask(void const* arg)
 {
     //One time osDelayUntil initialization
+    static const uint32_t BLUE_LED_STATUS_STDID = 0xCCC;
+    static const uint32_t BLUE_LED_TOGGLE_FREQ = 100;
```
JessieG123, line 7: This number here is wrong too. Please see what I said about the green one.

JessieG123 (changes requested):

Recruit-Training/Embedded-Recruit-Training/Core/Src/BlueLedToggleTask.c
```diff
@@ -11,7 +13,32 @@ void blueLedToggleTask(void const* arg)
     {
         //TODO: Add BLUE_LED_TOGGLE_FREQ to prevWakeTime
         osDelayUntil(prevWakeTime);
+        prevWakeTime += BLUE_LED_TOGGLE_FREQ;
```
duan-le, line 16: This should be before `osDelayUntil(prevWakeTime);`

Recruit-Training/Embedded-Recruit-Training/Core/Src/GreenLedToggleTask.c
```diff
@@ -11,7 +17,28 @@ void greenLedToggleTask(void const* arg)
     {
         //TODO: Add GREEN_LED_TOGGLE_FREQ to prevWakeTime
         osDelayUntil(prevWakeTime);
+        prevWakeTime += GREEN_LED_TOGGLE_FREQ;
```
duan-le, line 20: This should also be before `osDelayUntil(prevWakeTime);`

duan-le:
https://github.com/UCSolarCarTeam/Development-Resources/tree/master/Development-Guide/Style-Guide

You can take a look for naming conventions.

Recruit-Training/Embedded-Recruit-Training/Core/Src/main.c
```diff
@@ -328,41 +328,42 @@ void HAL_CAN_RxFifo0MsgPendingCallback(CAN_HandleTypeDef* hcan)
...
         {
-            int isGreen = 1;
-            for (int i = 2; i < 8; i++)
+            BlueToggleFlag = 0;
+        }
+        
+         if (hdr.StdId == GREEN_MESSAGE_STDID && hdr.DLC == 1)
+        {
+            uint8_t mask = 0b00000011
+            if (data[0] & mask == mask)
```
badeniran, line 353: This is technically correct, but the instructions explicitly say that the upper 6 bits are **not** set. The way you currently have it, if `data[0] = 0b11111111`, then it would still evaluate to true. A little hint is that you don't need a mask for this one

Recruit-Training/Embedded-Recruit-Training/Core/Src/main.c
```diff
@@ -328,41 +328,42 @@ void HAL_CAN_RxFifo0MsgPendingCallback(CAN_HandleTypeDef* hcan)
...
+                BlueToggleFlag = 0;
+            }
         }
-    
-        if (data[0] == 1 && data[1] == 1)
+        else
         {
-            int isGreen = 1;
-            for (int i = 2; i < 8; i++)
+            BlueToggleFlag = 0;
```
badeniran: I would say you don't need this else statement. We only want to update the toggle flag if the message has the blue led message.

Recruit-Training/Embedded-Recruit-Training/Core/Src/main.c
```diff
@@ -328,41 +328,42 @@ void HAL_CAN_RxFifo0MsgPendingCallback(CAN_HandleTypeDef* hcan)
...
-                    GreenToggleFlag = 1;
-
+                GreenToggleFlag = 0;
             }
         }
-    }
+        else
+        {
+            GreenToggleFlag = 0;
+        }
```
badeniran, line 365: I would say you don't need this else statement. We only want to update the toggle flag if the message has the green led message.

Recruit-Training/Embedded-Recruit-Training/Core/Src/main.c
badeniran, line 353: This sort of works, but the instructions explicitly state that for the green message, the upper 6 bits are **not** set.
If `data[0] == 0b11111111`, the if statement would evaluate to true, which we don't want. 
A little hint is that you don't need a mask for this condition.

badeniran (changes requested):
A lot cleaner, just a couple fixes and you should be good!

badeniran (approved):
Looks good to me! All that's left is to see if it works haha. Keep the pull request open till then

PR #171, opened 2020-11-07
Embedded Teaser Done

Ahmed-I-Abdullah:
Embedded Teaser Training Done. 
The following are the conditions for the motor to be valid:
1- Power on, positive velocity, forward direction
2- Power on, negative velocity, reverse direction
3- Power is off and velocity must be zero but direction can be either reverse or forward
4- Power on and direction bit of velocity is opposite to direction of motor but velocity is zero ( since we are using sign/magnitude representation of binary numbers which has two zeros).

The following are the conditions for the lights to be valid:
1- Only one option is chosen for the headlights
2- If hazards are off, only one or none of the signals can be on
3- If hazards are on, both signals must be on.


Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -2,5 +2,86 @@
...
+	uint8_t maskedHeadLights = testLight & 0b00000111;
+	uint8_t maskedSignals = testLight & 0b00011000;
+	
+	if(!(maskedHeadLights && !(maskedHeadLights & (maskedHeadLights-1))) || !(maskedSignals && !(maskedSignals & (maskedSignals-1)))) // if more than 1 headlight state is chosen or more than one signal is on return 0
+		return 0;
+	
+	return 1;
+}
 
 }
```
duan-le, line 62: You have an extra `}` here.

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -2,5 +2,86 @@
...
+	
+	if ((testMotor & motorMask) == 0b10000000) // check the case that motor is in reverse (power on, negative velocity and reverse)
+		return 1;
+	else if ((testMotor & motorMask) == motorMask) // check the case that motor is in forward (power on, positive velocity and forward)
+		return 2;
+	else if (testMotor == 0) // check case that power is off
+		return 3;
+	
+	return 0;
+}
```
duan-le, line 76: You already have a definition for `int motorCheck(uint8_t testMotor)` starting on line `37`.

duan-le (changes requested):

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -2,5 +2,86 @@
...
+int lightsCheck (uint8_t testLight)
+{
+	uint8_t maskedHeadLights = testLight & 0b00000111;
+	uint8_t maskedSignals = testLight & 0b00011000;
+	
+	if(!(maskedHeadLights && !(maskedHeadLights & (maskedHeadLights-1))) || !(maskedSignals && !(maskedSignals & (maskedSignals-1)))) // if more than 1 headlight state is chosen or more than one signal is on return 0
+		return 0;
+	
+	return 1;
+}
```
duan-le, line 87: You also have a definition for `int lightsCheck (uint8_t testLight)` starting on line `51`.

duan-le (changes requested):

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
Ahmed-I-Abdullah: oh, I have no clue what happened there. Thanks!

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -2,5 +2,59 @@
...
+		for (int i = 0; i < sizeof outputArray / sizeof *outputArray; i++)
+		{
+			outputArray[i] = *(data + i);
+		}	
+	}
+	else
+	{
+		for (int i = 0; i < sizeof outputArray / sizeof *outputArray; i++)
+		{
+			outputArray[i] = 0;
```
badeniran, line 18: If only the lights is valid, the outputArray[2] should match the data[2], similarly, if only the motors are valid, outputArray[0] and outputArray[1] should match data[0] and data[1] respectively

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -2,5 +2,59 @@
 
 void trainingTask(uint8_t* data)
 {
+	validData = 0;
+	
+	if (motorCheck(*data) && motorCheck(*data) == motorCheck(*(data + 1)) && lightsCheck(*(data + 2)) ) // set outputArray if motors data are valid, motors are in sync and lights data is valid
```
badeniran, line 7: I would say it's a better idea to store the return values of these functions in variables. That way you don't need to call the functions multiple times since you call them again in the set of if statements at the bottom of the function

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -2,5 +2,59 @@
...
+	{
+		validData = validData | (1 << LIGHTSBOARD);
+	}	
+}
+
+int motorCheck (uint8_t testMotor)
+{
+	uint8_t motorMask = 0b11000001;
+	
+	if ((testMotor & motorMask) == 0b10000000) // check the case that motor is in reverse (power on, negative velocity and reverse)
```
badeniran, line 41: If the sign of the velocity is 1, the velocity is negative, if the sign of the velocity is 0, the velocity is positive

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -2,5 +2,59 @@
...
+		return 1;
+	else if ((testMotor & motorMask) == motorMask) // check the case that motor is in forward (power on, positive velocity and forward)
```
badeniran, line 43: Same as above

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -2,5 +2,59 @@
...
+		return 2;
+	else if (testMotor == 0) // check case that power is off
```
badeniran, line 45: It's possible for the motor to be in the "forward" direction with the motor being off, so you'll need to account for that

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -2,5 +2,59 @@
...
+	
+	return 0;
+}	
 
+int lightsCheck (uint8_t testLight)
+{
+	uint8_t maskedHeadLights = testLight & 0b00000111;
+	uint8_t maskedSignals = testLight & 0b00011000;
+	
+	if(!(maskedHeadLights && !(maskedHeadLights & (maskedHeadLights-1))) || !(maskedSignals && !(maskedSignals & (maskedSignals-1)))) // if more than 1 headlight state is chosen or more than one signal is on return 0
```
badeniran, line 56: Would be a good idea to split this if statement and perhaps assign the boolean conditions to variables. It makes it easier to read

badeniran: For the signals, you should have a case where if the hazards are on, both signals should be on

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -2,5 +2,59 @@
 
 void trainingTask(uint8_t* data)
 {
+	validData = 0;
+	
+	if (motorCheck(*data) && motorCheck(*data) == motorCheck(*(data + 1)) && lightsCheck(*(data + 2)) ) // set outputArray if motors data are valid, motors are in sync and lights data is valid
+	{
+		for (int i = 0; i < sizeof outputArray / sizeof *outputArray; i++)
```
badeniran, line 9: Do `sizeof(outputArray)/sizeof(*outputArray)`. You could also just hardcode it to 3

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -2,5 +2,59 @@
...
+		{
+			outputArray[i] = *(data + i);
+		}	
+	}
+	else
+	{
+		for (int i = 0; i < sizeof outputArray / sizeof *outputArray; i++)
```
badeniran, line 16: Same as above

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Test/TrainingTest.c
```diff
@@ -4,25 +4,48 @@
...
+	RUN_TEST(test_OnlyMotorsInvalid);
 }
 
 void test_EverythingValid()
 {
-
+	printf("\nTesting trainingTask - Should Pass\n");
+	uint8_t inputData[3] = {0b11001011, 0b11010111, 0b00001010}; 
+	trainingTask (inputData);
+	TEST_ASSERT_EQUAL_UINT8_ARRAY (inputData, outputArray, 3);
```
badeniran, line 18: Not really necessary, but it would be a good idea to use the `_MESSAGE` versions of the test macro. The only difference is that there's one more argument that is a string for the message to be printed in case of a failure

badeniran (commented):
Good job so far! Just a few logical fixes and some changes to make the code a bit more readable

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -1,49 +1,61 @@
...
+	if (motorOneState)  // set first motor bit in validData to 1 if motor's data is valid
 	{
 		validData = validData | (1 << FIRSTMOTOR);
 	}
 	
-	if (motorCheck (*(data + 1))) // set second motor bit in validData to 1 if motor's data is valid
+	if (motorTwoState) // set second motor bit in validData to 1 if motor's data is valid
 	{
 		validData = validData | (1 << SECONDMOTOR);
 	}
```
badeniran: Remember, if one motor is invalid, both motors are invalid

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -1,49 +1,61 @@
...
 {
-	uint8_t motorMask = 0b11000001;
+	uint8_t forwardMask = 0b10000001;
+	uint8_t reverseMask = 0b11000000;
 	
-	if ((testMotor & motorMask) == 0b10000000) // check the case that motor is in reverse (power on, negative velocity and reverse)
+	if ((testMotor & forwardMask) == forwardMask) // check the case that motor is in reverse (power on, negative velocity and reverse)
 		return 1;
-	else if ((testMotor & motorMask) == motorMask) // check the case that motor is in forward (power on, positive velocity and forward)
+	else if ((testMotor & reverseMask) == reverseMask) // check the case that motor is in forward (power on, positive velocity and forward)
```
badeniran: Logically, this doesn't work for all cases. For the first if statement, if the velocity sign is 1, when you do the AND with the mask, the resulting bit 7 will go to zero and it will appear that everything is fine. For the second if statement, if the forward/reverse is a 1, when you do the AND with the mask, the resulting bit 0 will go to zero and it will appear that everything is fine.

badeniran: A little hint: When you create a mask, you turn on the bits you are interested in. In your case, for knowing the velocity sign, the power state, and the forward/reverse, you are interested in bits 0,6, and 7. After ANDing with the mask, you compare the result to the desired values post AND. So you would compare to your current `forwardMask` for the forward case and `reverseMask` for the reverse case

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -1,49 +1,61 @@
...
 		return 2;
-	else if (testMotor == 0) // check case that power is off
+	else if ((testMotor & 0b00000001) == 0b00000001) // check case that power is off and motor on "forward"
```
badeniran, line 55: Similar to above. The power bit could be on and things would still be fine

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -52,8 +64,12 @@ int lightsCheck (uint8_t testLight)
 {
 	uint8_t maskedHeadLights = testLight & 0b00000111;
 	uint8_t maskedSignals = testLight & 0b00011000;
+	uint8_t maskedHazards = testLight & 0b00111000;
+	uint8_t hazardsOn = testLight & 0b00100000;
+	int oneHeadlightOn = maskedHeadLights && !(maskedHeadLights & (maskedHeadLights-1));
+	int oneSignalOn = maskedSignals && !(maskedSignals & (maskedSignals-1));
 	
-	if(!(maskedHeadLights && !(maskedHeadLights & (maskedHeadLights-1))) || !(maskedSignals && !(maskedSignals & (maskedSignals-1)))) // if more than 1 headlight state is chosen or more than one signal is on return 0
+	if(!oneHeadlightOn || (hazardsOn && maskedHazards != 0b00111000) || (!hazardsOn && !oneSignalOn)) // if more than 1 headlight state is chosen or both signals are not on while hazard is on or more than one signal is on return 0
```
badeniran, line 72: If none of the signals lights are on, you'll return a 0, but it's valid for neither of the signal lights to be on

badeniran (changes requested):
A couple more logical issues

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -21,38 +23,23 @@ void trainingTask (uint8_t* data)
...
 {
-	uint8_t forwardMask = 0b10000001;
-	uint8_t reverseMask = 0b11000000;
+	uint8_t motorMask = 0b11000001;
 	
-	if ((testMotor & forwardMask) == forwardMask) // check the case that motor is in reverse (power on, negative velocity and reverse)
+	if ((testMotor & motorMask) == 0b10000001) // check the case that motor is in reverse (power on, negative velocity and reverse)
 		return 1;
-	else if ((testMotor & reverseMask) == reverseMask) // check the case that motor is in forward (power on, positive velocity and forward)
+	else if ((testMotor & motorMask) == 0b11000000) // check the case that motor is in forward (power on, positive velocity and forward)
```
badeniran: I think the comments for these if statements are flipped

badeniran (changes requested):
Looks good! Just one more thing regarding comments

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -1,6 +1,63 @@
...
+	uint8_t motorMask = 0b11000001;
+	
+	if ((testMotor & motorMask) == 0b10000001) // check the case that motor is in reverse (power on, negative velocity and reverse)
+		return 1;
+	else if ((testMotor & motorMask) == 0b11000000) // check the case that motor is in forward (power on, positive velocity and forward)
+		return 2;
+	else if ((testMotor & 0b10000001) == 0b00000001) // check case that power is off and motor on "forward"
+		return 3;
+	else if (testMotor == 0) // check case that power is off
+		return 4;
```
badeniran, line 45: One thing I just remembered. It's possible that the velocity is 0 while the car is on and has a direction. Unfortunately, we've chosen sign-mag for representing the velocity, so 0 has 2 representations: 0b100000 and 0b000000. So it's technically not incorrect for the direction of the car to be forward, and the sign of the velocity to be 1 **if** the magnitude of the velocity is 0.

Another point I forgot to point out is that if the motors are off, the velocity should be 0. Then again you can argue that the motors could've recently been on and just turned off so because of inertia, the velocity isn't 0. I'm fine with  it either way, but you'll just have to explicitly state the latter assumption.

Ahmed-I-Abdullah: Can the direction also be reverse while the power is off? If so I can make one if statement to check this statement and that v=0 when the motor is off.

badeniran: Yes, that's also valid

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Test/TrainingTest.c
```diff
@@ -4,25 +4,47 @@
...
 
 void test_OnlyLightsInvalid()
 {
-
+	printf("\nTesting trainingTask - Should Pass\n");
+	uint8_t inputData[3] = {0b10101011, 0b10101101, 0b00110010};
+	uint8_t expectedOutput[3] = {0b10101011, 0b10101101, 0};
+	trainingTask (inputData);
+	TEST_ASSERT_EQUAL_UINT8_ARRAY_MESSAGE (expectedOutput, outputArray, 3, "Only motors data should be copied into outputArray as these are valid data while the lights data is invlaid");
+	TEST_ASSERT_EQUAL_UINT8_MESSAGE (3, validData, "Bits 0 and 1 of validData should be set to 1 as they represent the motors which have valid inputs");
```
JessieG123, line 39: Your messages are slightly wordy. It should be short, clear and concise.

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Test/TrainingTest.c
```diff
@@ -4,25 +4,47 @@
...
+	uint8_t expectedOutput[3] = {0};
+	trainingTask (inputData);
+	TEST_ASSERT_EQUAL_UINT8_ARRAY_MESSAGE (expectedOutput, outputArray, 3, "outputArray should have all zeros as none of the data inputs is valid");
+	TEST_ASSERT_EQUAL_UINT8_MESSAGE (0, validData, "The last 3 bits of validData should remain zero as all data is invalid");
 }
 
 void test_OnlyLightsInvalid()
 {
-
+	printf("\nTesting trainingTask - Should Pass\n");
```
JessieG123, line 34: There's no need to have a print statement like that in the test.

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
```diff
@@ -1,6 +1,63 @@
...
 {
+	validData = 0;
+	int motorOneState = motorCheck (*data);
+	int motorTwoState = motorCheck (*(data + 1));
+	int lightState = lightsCheck (*(data + 2));
+	
+	if (motorOneState && motorOneState == motorTwoState) // set outputArray values of motor if motors data are valid, motors are in sync
+	{
+		outputArray[0] = *data;
+		outputArray[1] = *(data + 1);
```
JessieG123, line 13: Should write in array notation instead to be consistent and it is also what we use.

JessieG123 (changes requested):

Recruit-Training/Advanced-Recruit-Training/Embedded-Teaser/Training/Core/Src/Training.c
Ahmed-I-Abdullah: What's the array notation

duan-le (approved):
Looks good to me. Good job!

JessieG123 (approved):
Looks awesome, great job! You can now close the PR.

PR #186, opened 2021-01-09
Finished VisComm Teaser

MarceloLiGonzales:
I was not sure what to do for the add Member button so I used a simple text input prompt.


Hesham-E:
Awesome work here Marcelo! There's only one thing that I found for you to improve. It's nice to ensure that the user will end up on the first page on the stacked pages. There's two ways you can do this. 1. You can code it into the constructor of MainWindow. 2. This one almost feels like cheating, but you can select the first page in MainWindow.ui and it will also work.

DatLam25:
Good work, everything seems fine to me.

DatLam25 (approved):
Wait for another approval. When you're done with this training, remember to _**not**_ merge but just delete the PR and the branch. Good work!

PR #231, opened 2021-10-23
User/tyler/recruit training

ttran1017:
Completed pointer training


Recruit-Training/Basic-Recruit-Training/person/Person.cpp
```diff
@@ -0,0 +1,54 @@
...
+    int combined_age = personArray[0]->getAge();
+    for(int i = 1; i < size; i++)
+        combined_age+=personArray[i]->getAge();
+    return combined_age;
+}
+
+void Person::birthday(Person& x)
+{
+    x.setAge(x.getAge()+1);
+}
```
JessieG123, line 54: It is best practice to have a new line after the end of the file for C/C++. 

More info: https://stackoverflow.com/questions/72271/no-newline-at-end-of-file-compiler-warning

Recruit-Training/Basic-Recruit-Training/pointers/main.cpp
```diff
@@ -55,3 +60,10 @@ void var(int *a, int b)
     *a = 365;
     b = 912;
 }
+
+void mySwap(int* a, int* b)
+{
+    *a = *a xor *b;
+    *b = *b xor *a;
+    *a = *a xor *b;
```
JessieG123, line 68: This is an interesting way to go about it and does work. I've never seen anyone do swaps like this but a quick google search tells me that this is cool but bad practice.  It may be more efficient if you were programming in assembly but modern CPUs can usually handle swaps in a single instruction. The compile might not notice what you're doing and this might end up being slower (3 instructions vs 1). Also, if you were to debug this in half a year that's probably painful. 

More info why this is might be a bad idea:
https://en.wikipedia.org/wiki/XOR_swap_algorithm

JessieG123 (changes requested):

Recruit-Training/Basic-Recruit-Training/person/Person.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+const std::string& Person::getName() const
+{
+    return name_;
+}
+
+void Person::setName(const std::string& newName)
+{
+    name_=newName;
```
JessieG123, line 26: Please add a space between operators.

JessieG123 (changes requested):

Recruit-Training/Basic-Recruit-Training/person/Person.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+void Person::setName(const std::string& newName)
+{
+    name_ = newName;
+}
+
+void Person::setAge(int newAge)
+{
+    delete age_;
+    age_ = new int(newAge);
```
JessieG123, line 32: This probably works but you can just dereference age_ and assign it to newAge. When you allocate memory for a variable you can change the value stored in that memory address. If you call setAge a bunch of time, constantly deleting and re allocating is likely going to be pretty expensive.

JessieG123 (changes requested):

disneyonbryce (approved):
Looks good!

JessieG123 (approved):
Looks great! Please **close** this PR and **delete** your branch! 
sealofapproval

PR #233, opened 2021-10-23
Updated installation instructions.

JaydenBischoff:


JaydenBischoff:
@liconway What error do you get from that command? I tried it in Linux and Windows and I didn't get any.

JensVarughese (approved):
Looks good to me!

Dillonsd:
> @liconway What error do you get from that command? I tried it in Linux and Windows and I didn't get any.

I think global npm installs do require superuser privileges. Not sure if it's a good idea to encourage that though, I think npm install runs some post/pre-install scripts that could potentially be harmful. I think we talked about using nvm right?

JaydenBischoff:
> > @liconway What error do you get from that command? I tried it in Linux and Windows and I didn't get any.
> 
> I think global npm installs do require superuser privileges. Not sure if it's a good idea to encourage that though, I think npm install runs some post/pre-install scripts that could potentially be harmful. I think we talked about using nvm right?

Yes we are using nvm for installing node. I will have to see if you can install angular using nvm but in the case that you can't, would `npm install @angular/cli` work then?

PR #239, opened 2021-11-05
completed pointers

haley-lynn:


JessieG123 (commented):
Usually we only open PRs when everything is done, in this case you still need to do the person training because it's part of the whole basic training. Just letting you know about this, for this time it's alright just continue to push to this branch and the PR will automatically update with changes.

Recruit-Training/Basic-Recruit-Training/person/Person.cpp
```diff
@@ -0,0 +1,56 @@
...
+  {
+      age_ = new int[1];
+  }
+
+  *age_ = newAge;
+}
+void Person::printInfo() const 
+{
+  printf("%s", name_.c_str());
+  printf("%d", *age_);
```
JessieG123, line 42: If you didn't know, this is the C way of printing things out. In C++ it looks a bit like `std::cout << "whatever" << std::endl;`.  It both works but there are some differences that you can read about in here: https://stackoverflow.com/questions/2872543/printf-vs-cout-in-c.

Recruit-Training/Basic-Recruit-Training/person/Person.cpp
```diff
@@ -0,0 +1,56 @@
+#include "Person.h"
+#include "main.cpp"
```
JessieG123, line 2: Do you need this?

JessieG123 (changes requested):

PR #242, opened 2021-11-20
Finish Basic Recruit Training

ri-yad-a:
Finished Basic Recruit Training

### DIFF
--- a/Recruit-Training/Basic-Recruit-Training/person/Person.cpp
+++ b/Recruit-Training/Basic-Recruit-Training/person/Person.cpp
@@ -1,43 +1,37 @@
 #include <iostream> 
 #include "Person.h"
+#include <string>
 
-
-
-
-std::string(n);
+std::string n;
 int a;
 
 Person::Person() 
 {
-
+	name_ = "";
+	age_ = new int(0);
 }
 
 Person::~Person() 
 {
-
+	free(age_);
 }
 
 Person::Person(const std::string& name, int age) 
 {
 	name_ = name;
 	age_ = new int(age);
-
 } 
 
 // getters and setters
 int Person::getAge() const 
 { 
-
 	return *age_;
-
 }
 
 
 const std::string& Person::getName() const 
 {
-
 	return name_;
-
 }
 
 void Person::setName(const std::string& newName)
@@ -48,32 +42,24 @@ void Person::setName(const std::string& newName)
 	
 void Person::setAge(int newAge) 
 {
-
 	*age_ = newAge;
-
 }
 
 // Print out the name and age of the person
 void Person::printInfo() const 
 {
-
 	std::cout << "Name: " << name_ << ", Age: " << *age_ << std::endl;
-
 }
 
 // Print the combined age of everyone pointed to by personArray
 int Person::combinedAge(Person** personArray, int size) 
 {
-
 	int ageTotal = 0;
 	for (int i = 0; i < size; i++) {
 		ageTotal = ageTotal + personArray[i]->getAge();
 	}
 
 	return ageTotal;
-
-
-
 }
 
 // Increases the Person's Age by 1
@@ -81,6 +67,3 @@ void Person::birthday(Person& x)
 {
 	x.setAge(x.getAge() + 1);
 }
-
-
-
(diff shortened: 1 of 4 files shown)


Recruit-Training/Basic-Recruit-Training/person/Person.cpp
```diff
@@ -0,0 +1,86 @@
+#include <iostream> 
+#include "Person.h"
+
+
+
+
```
JessieG123, line 6: You can probably remove these excess spaces

Recruit-Training/Basic-Recruit-Training/person/Person.cpp
```diff
@@ -0,0 +1,86 @@
...
+} 
+
+// getters and setters
+int Person::getAge() const 
+{ 
+
+	return *age_;
+
+}
```
JessieG123, line 33: You have a lot of spaces. I think the return is also indented more than 4 spaces. Please be careful of spacing.

Recruit-Training/Basic-Recruit-Training/person/Person.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+// Increases the Person's Age by 1
+void Person::birthday(Person& x)
+{
+	x.setAge(x.getAge() + 1);
+}
+
+
+
```
JessieG123, line 86: Just leave one extra line at the end.

JessieG123 (changes requested):

Recruit-Training/Basic-Recruit-Training/person/Person.cpp
```diff
@@ -0,0 +1,86 @@
...
+Person::Person() 
+{
+
+}
+
+Person::~Person() 
+{
+
+}
```
JessieG123, line 18: Hint: something goes in these functions. 
Person() is a constructor so something like `this.name = "name" ` should be in there. 
~Person() is a destructor so it is used to delete a pre-allocated variable. Hint: it might be age.

Recruit-Training/Basic-Recruit-Training/person/Person.cpp
```diff
@@ -0,0 +1,86 @@
+#include <iostream> 
+#include "Person.h"
+
+
+
+
+std::string(n);
```
JessieG123, line 7: If you want to make a string variable in C++, syntax is like `std::string n;` similar to when you make a `int` or `char` variable just that for string it just need the `std::` because it is part of the library. I think you also need to add `#include <string>` to your includes.

JessieG123 (changes requested):

Recruit-Training/Basic-Recruit-Training/person/Person.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+Person::Person() 
+{
+	name_ = "";
+	age_ = new int(0);
+}
+
+Person::~Person() 
+{
+	free(age_);
```
DatLam25, line 16: It's better to use delete here. [Here's an article talking about the difference](https://www.geeksforgeeks.org/new-vs-malloc-and-free-vs-delete-in-c/)

Recruit-Training/Basic-Recruit-Training/person/Person.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+	std::cout << "Name: " << name_ << ", Age: " << *age_ << std::endl;
+}
+
+// Print the combined age of everyone pointed to by personArray
+int Person::combinedAge(Person** personArray, int size) 
+{
+	int ageTotal = 0;
+	for (int i = 0; i < size; i++) {
+		ageTotal = ageTotal + personArray[i]->getAge();
```
DatLam25, line 59: You can use the `+=` instead of `=`

Recruit-Training/Basic-Recruit-Training/person/Person.cpp
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	return ageTotal;
+}
+
+// Increases the Person's Age by 1
+void Person::birthday(Person& x)
+{
+	x.setAge(x.getAge() + 1);
```
DatLam25, line 68: Since birthday is in the Person class, you can just use the age_ directly here.

PR #244, opened 2021-12-05
telemetry advanced recruit training

jjason-nn:


JaydenBischoff (approved):
Looks good. I think that for loops could make your code shorter as an improvement but this is not necessary for the training.

Dillonsd (approved):
Looks good. For future reference, don't add the package and package-lock jsons unless you're making changes to dependencies (like adding a new npm module). Good job!

PR #261, opened 2022-03-19
User/haley/telemetry training

haley-lynn:
Completed Telemetry Training


JaydenBischoff (approved):
Looks good. Spacing is a bit off but that doesn't need to be fixed for the training

PR #328, opened 2023-11-04
User/johnnytran/basictraining

jtran217:
Telemetry should be in there now


Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -2,24 +2,66 @@ import BatteryInput from "~/components/batteryInput";
...
 import SpeedInput from "~/components/speedInput";
 import WeatherInput from "~/components/weatherInput";
+import Button from "~/components/Button/Button.tsx";
+import {useState} from "react";
 
 const App = () => {
-  const calculateRange = () => {
-    return;
+  const [speed,setSpeed] = useState(null);
+  const [batteryPercent, changeBatteryPercent] = useState(0);
```
brian-ngyn, line 10: Consistent naming, it should be setBatteryPercent instead of changeBatteryPercent

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -2,24 +2,66 @@ import BatteryInput from "~/components/batteryInput";
...
+const batterChangeHandler = (e) => {
+    if(e.target.value > 100 || e.target.value<0) {
+      setBatteryValid(false);
+    }
+    else{
+      setBatteryValid(true);
+      changeBatteryPercent(e.target.value);
+      console.log("Battery %" + batteryPercent)}
+
+}
```
brian-ngyn, line 45: Can you think of a way of handling the valid/invalid state of the speed and battery without a separate state variable?

Also, do we _**need**_ to store the range in a state variable as well? Can we derive the value without using state?

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -2,24 +2,66 @@ import BatteryInput from "~/components/batteryInput";
...
 import WeatherInput from "~/components/weatherInput";
+import Button from "~/components/Button/Button.tsx";
+import {useState} from "react";
 
 const App = () => {
-  const calculateRange = () => {
-    return;
+  const [speed,setSpeed] = useState(null);
+  const [batteryPercent, changeBatteryPercent] = useState(0);
+  const [weather, weatherChange] = useState(0);
```
brian-ngyn, line 11: Consistent naming, it should be setWeather instead of weatherChange

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/package-lock.json
brian-ngyn, line 1: Package-lock is only when you use npm, did you use npm instead of yarn by accident?

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -2,24 +2,66 @@ import BatteryInput from "~/components/batteryInput";
...
+  const[speedValid,setSpeedValid]= useState(false);
+  const[batteryValid,setBatteryValid]= useState(false);
+  const [calc, setCalc] = useState(0)
+  const [initialCalc, setInitialCalc] =useState(false);
+
+  const calculateRange = (e) => {
```
brian-ngyn, line 17: This function can still be called even though inputs are valid?

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -2,24 +2,66 @@ import BatteryInput from "~/components/batteryInput";
...
+    setInitialCalc(true);
+  }
+  const speedChangeHandler = (e) => {
+    if(e.target.value > 90 || e.target.value < 0){
+      setSpeedValid(false)
+    }
+    else{
+      setSpeedValid(true)
+      setSpeed(e.target.value)
+      console.log(speed)
```
brian-ngyn, line 29: Remove console.log

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -2,24 +2,66 @@ import BatteryInput from "~/components/batteryInput";
...
+    }
+}
+  const weatherChangeHandler = (e) => {
+    weatherChange(e.target.value);
+    console.log(weather)
```
brian-ngyn, line 34: Remove console.log

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -2,24 +2,66 @@ import BatteryInput from "~/components/batteryInput";
...
+  const [calc, setCalc] = useState(0)
+  const [initialCalc, setInitialCalc] =useState(false);
+
+  const calculateRange = (e) => {
+    e.preventDefault();
+    setCalc(-(speed * speed * batteryPercent / 2500) + (4 * batteryPercent) + weather);
+    setInitialCalc(true);
+  }
+  const speedChangeHandler = (e) => {
+    if(e.target.value > 90 || e.target.value < 0){
```
brian-ngyn, line 23: We should be validating if the input is a number or not, I was able to type in "e" and "-" into the input box

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -2,24 +2,66 @@ import BatteryInput from "~/components/batteryInput";
...
+      setSpeed(e.target.value)
+      console.log(speed)
+    }
+}
+  const weatherChangeHandler = (e) => {
+    weatherChange(e.target.value);
+    console.log(weather)
   }
+const batterChangeHandler = (e) => {
+    if(e.target.value > 100 || e.target.value<0) {
```
brian-ngyn, line 37: We should be validating if the input is a number or not, I was able to type in "e" and "-" into the input box. This is because the number input can accept floating-point numbers (e)

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/components/batteryInput/index.tsx
```diff
@@ -1,8 +1,9 @@
-const BatteryInput = () => {
+const BatteryInput = (props) => {
   return (
     <div className="flex w-full flex-col items-center gap-2">
       <label>Battery Percentage (%):</label>
```
brian-ngyn, line 4: In App.tsx, you give the battery an intial value of 0, but when the app mounts for the first time, there is no initial value for the battery input. 

Why is this? Look into controlled input fields work in React

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/components/speedInput/index.tsx
```diff
@@ -1,9 +1,10 @@
-const SpeedInput = () => {
+const SpeedInput = (props) => {
   return (
     <>
       <div className="flex w-full flex-col items-center gap-2">
         <label>Speed (km/h):</label>
         <input
```
brian-ngyn, line 6: In App.tsx, you give the battery an intial value of null, but the input field is of type "number". This can cause a type error in the future.

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -2,24 +2,66 @@ import BatteryInput from "~/components/batteryInput";
...
+import {useState} from "react";
 
 const App = () => {
-  const calculateRange = () => {
-    return;
+  const [speed,setSpeed] = useState(null);
+  const [batteryPercent, changeBatteryPercent] = useState(0);
+  const [weather, weatherChange] = useState(0);
+  const[speedValid,setSpeedValid]= useState(false);
+  const[batteryValid,setBatteryValid]= useState(false);
```
brian-ngyn, line 13: We're creating 2 state variables to check to see if the 2 inputs is valid. What if we had a form that had 50+ inputs, are we expecting to create 50+ corresponding valid states? There should be a better way to check if the inputs are valid without creating extra state.

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -2,24 +2,66 @@ import BatteryInput from "~/components/batteryInput";
...
+  const [calc, setCalc] = useState(0)
+  const [initialCalc, setInitialCalc] =useState(false);
```
brian-ngyn, line 15: Do we need to create extra state to store the calculated range? We should be able to derive the range from the 3 state variables `speed`, `batteryPercent`, `weather`.

Remember, when you update state in React, it will cause your whole component to re-render (unless you use useMemo or useCallback...will get to that in future lesson). Think about using this to your advantage and you may not need extra state to store the range.

PR #339, opened 2024-05-15
Feat: Weather, Speed, Battery, and Range implementation

David-Rodriguez-Barrios:


Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -1,25 +1,118 @@
...
 import BatteryInput from "~/components/batteryInput";
 import Header from "~/components/header";
 import SpeedInput from "~/components/speedInput";
 import WeatherInput from "~/components/weatherInput";
 
 const App = () => {
-  const calculateRange = () => {
-    return;
-  }
+  const [speed, setSpeed] = useState<number | undefined>(0);
```
brian-ngyn, line 9: speed, battery, and weather are all together. can we create a single state that represents all 3 of the inputs?

example const [inputs, setInputs] = useState({}), where the object is the key value pair for each of the 3 inputs?

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -1,25 +1,118 @@
...
+  const [speedValid, setSpeedValid] = useState(true);
```
brian-ngyn, line 10: we're using state to represent the valid state for battery and speed. do we need to create state for this?

remember that your app will re-render when any of the fields change, which is fine. But can we use that to our advantage and not create new state and make those "valid" variables a computed value from the state?

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -1,25 +1,118 @@
...
+  const [battery, setBattery] = useState<number | undefined>(100);
+  const [batteryValid, setBatteryValid] = useState(true);
+  const [weather, setWeather] = useState(50);
+  const [range, setRange] = useState<number | null>(null);
```
brian-ngyn, line 14: same thing for the range, do we need to use state for the range? we know what the formula is for the state, so is it possible to directly compute the range via the other state variables we defined?

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -1,25 +1,113 @@
...
+    } else {
+      setInputs({ ...inputs, speed: undefined });
+    }
+  };
+
+  const sliderChange = (num: string) => {
+    const value = parseFloat(num);
+    setInputs({ ...inputs, weather: value });
+    setVisible(false);
+  };
```
brian-ngyn, line 50: in all 3 of these functions, you are doing the same thing,     setVisible(false);

can you leverage useReducer and create 3 actions which will update each of the respective inputs?

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -1,25 +1,113 @@
...
+          </button>
+          {visible &&
+            inputs["speed"] !== undefined &&
+            inputs["battery"] !== undefined &&
+            inputs["weather"] !== undefined &&
+            inputs["battery"] <= 100 &&
+            inputs["battery"] >= 0 &&
+            inputs["speed"] <= 100 &&
+            inputs["speed"] >= 0 && (
+              <p className="mt-10">
```
brian-ngyn, line 98: you're doing a lot of recomputation here, can you derive the valid and invalid states of each input field and reuse that variable?

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -1,25 +1,148 @@
...
+    weather: 50,
+  };
+
   const calculateRange = () => {
-    return;
+    event?.preventDefault();
+    setVisible(true);
+  };
+  function BatteryValid() {
+    const battery = inputs["battery"];
```
brian-ngyn, line 34: can you try to simplify the if else statement to a single ternary?

Recruit-Training/Advanced-Recruit-Training/Telemetry-Teaser/src/App.tsx
```diff
@@ -1,25 +1,148 @@
...
+    if (battery !== undefined && battery <= 100 && battery >= 0) {
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  function SpeedValid() {
+    const speed = inputs["speed"];
+    if (speed !== undefined && speed <= 90 && speed >= 0) {
```
brian-ngyn, line 44: can you try to simplify the if else statement to a single ternary?

